PR #4, opened 2016-02-02
Combine Brooklyn docs into single page/pdf - for review

drigodwin:
Additions to templates and build scripts to build a single page version of the documents and PDF
* Requires wkhtmltopdf

### DIFF
--- a/_build/buildPDF.sh
+++ b/_build/buildPDF.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+#
+# this generates A PDF from a HTML file
+
+if [ ! -x _build/build.sh ] ; then
+  echo "ERROR: script must be run in root of docs dir"
+  exit 1
+fi
+
+if [ "$#" -lt 2 ] ; then
+	echo "Usage: buildPDF.sh <source html> <target PDF>"
+	exit 1
+fi
+
+command -v wkhtmltopdf >/dev/null 2>&1 || { 
+	echo "ERROR: script requires wkhtmltopdf but it's not installed." >&2; exit 1;
+	}
+	
+# run the PDF build with wkhtmltopdf
+wkhtmltopdf --page-size Letter --margin-top 0.75in --margin-right 0.75in --margin-bottom 0.75in --margin-left 0.75in --encoding UTF-8 $1 $2
+if [ $? -eq 1 ] ; then
+	echo "Note that ContentNotFoundError is usually because of the invalid relational path of a local resource such as an image"
+else	
+	echo "PDF Built successfully."
+fi
--- a/_build/buildPDF.sh
+++ b/_build/buildPDF.sh
@@ -7,14 +7,15 @@ if [ ! -x _build/build.sh ] ; then
   exit 1
 fi
 
-if [ "$#" -lt 2 ] ; then
+if [ "$#" -ne 2 ] ; then
 	echo "Usage: buildPDF.sh <source html> <target PDF>"
 	exit 1
 fi
 
-command -v wkhtmltopdf >/dev/null 2>&1 || { 
-	echo "ERROR: script requires wkhtmltopdf but it's not installed." >&2; exit 1;
-	}
+if ! command -v wkhtmltopdf >/dev/null 2>&1; then
+	echo "ERROR: script requires wkhtmltopdf but it's not installed."
+	exit 1
+fi
 	
 # run the PDF build with wkhtmltopdf
 wkhtmltopdf --page-size Letter --margin-top 0.75in --margin-right 0.75in --margin-bottom 0.75in --margin-left 0.75in --encoding UTF-8 $1 $2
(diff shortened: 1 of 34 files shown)


aledsage:
Can you include in the README.md that `wkhtmltopdf` is needed, and say how to get it?

_build/buildPDF.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# this generates A PDF from a HTML file
+
+if [ ! -x _build/build.sh ] ; then
+  echo "ERROR: script must be run in root of docs dir"
```
alrick87: I'm not sure this is the correct error message after checking the file exists and executable. Possible solution to avoid this error is to get the absolute script path, i.e. ```$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )```

_build/buildPDF.sh
```diff
@@ -0,0 +1,27 @@
...
+  exit 1
+fi
+
+if [ "$#" -lt 2 ] ; then
```
alrick87: ```-lt``` should be ```-ne```, otherwise it will accept 3 or more args without complains

_build/buildPDF.sh
```diff
@@ -0,0 +1,27 @@
...
+	echo "Usage: buildPDF.sh <source html> <target PDF>"
+	exit 1
+fi
+
+command -v wkhtmltopdf >/dev/null 2>&1 || {
```
alrick87: In my opinion if clause would be more readable here:
```
if ! command -v wkhtmltopdf >/dev/null 2>&1; then
    echo "Error message"
    exit 1
fi

File: _build/buildPDF.sh [hunk 4]
```diff
@@ -0,0 +1,27 @@
...
+	echo "Usage: buildPDF.sh <source html> <target PDF>"
+	exit 1
+fi
+
+command -v wkhtmltopdf >/dev/null 2>&1 || { 
+	echo "ERROR: script requires wkhtmltopdf but it's not installed." >&2; exit 1;
+	}
+	
+# run the PDF build with wkhtmltopdf
+wkhtmltopdf --page-size Letter --margin-top 0.75in --margin-right 0.75in --margin-bottom 0.75in --margin-left 0.75in --encoding UTF-8 $1 $2
```
alrick87: maybe assign the ```wkhtmltopdf``` command to a variable like ```RUNPDF``` and then just do:
```
if ! $RUNPDF; then
    echo "Error message"
    exit 1
else
    echo "Success"
    exit 0
fi
```

zoneMergeStarted.html
```diff
@@ -0,0 +1,30 @@
...
+{% comment %}<!-- This page builds the single page getting started for the PDF, note that this name must be retained for conversion order -->{% endcomment %}
+
+	<div id="content_container" class="container">
+	<header>
+		<h1 id="content-top">Apache Brooklyn: Getting Started Guide</h1>
+	</header>
+		<h1>Contents</h1>
+	
+	    <nav><ul>
+		  {% assign visited = "" | split: "|" %}
```
alrick87: indentation issues here

zoneMergeManual.html
```diff
@@ -0,0 +1,31 @@
...
+{% comment %}<!-- This page builds the single page manual for the PDF, note that this name must be retained for conversion order -->{% endcomment %}
+
+	<div id="content_container" class="container">
+	<header>
+		<h1 id="content-top">Apache Brooklyn: User Manual</h1>
+	</header>
+		<h1>Contents</h1>
+		
+	    <nav><ul>
+		  {% assign visited = "" | split: "|" %}
```
alrick87: indentation issues here

style/css/singlePage.css
```diff
@@ -0,0 +1,56 @@
...
+	border: 0px solid #fff !important;
+	box-shadow: none;
+	padding-top: 0px;
+	padding-right: 0px;
+	padding-left: 0px;
+}
+pre, pre code{
+	overflow: auto;
+	white-space: pre-wrap;       /* CSS 3 */
+    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
```
alrick87: indentation

style/css/singlePage.css
```diff
@@ -0,0 +1,56 @@
...
+    white-space: -pre-wrap;      /* Opera 4-6 */
+    white-space: -o-pre-wrap;    /* Opera 7 */
+    word-wrap: break-word;       /* Internet Explorer 5.5+ */
+}
+.goToTop{
+	padding-right: 25px;
+    padding-top: 75px;
```
alrick87: indentation

_build/buildPDF.sh
drigodwin: This is the same as the start of build.sh, I think we should keep all scripts in a similar format if possible

aledsage:
Thanks @drigodwin - merging.

PR #5, opened 2016-02-02
Updates the getting started “policies” example.

aledsage:
Walks through auto-repair, auto-recover and auto-scaling of a simple tomcat cluster.

Note the "TODO"s:

* How do we make the yaml example easier to copy-paste? As with the example at http://brooklyn.apache.org/learnmore/blueprint-tour.html, there is no "copy to clipboard" action; and highlighting the text has the unfortunate effect of also copying he headings on the left (rather than just the yaml).

* Why doesn't the activity view show the `restart()` effector call made by the `ServerRestarter` policy? That part of the guide is commented out.

### DIFF
--- a/guide/start/policies.md
+++ b/guide/start/policies.md
@@ -9,169 +9,438 @@ children:
 - { section: "Managing with Policies", title_in_menu: "Policies"  } 
 ---
 
-
-
 ## A Clustered Example
 
-We'll now look at a more complex example that better shows the capabilities of Brooklyn. 
+We'll now look at a more complex example that better shows the capabilities of Brooklyn, including
+management of a running clustered application.
 
-We'll start by deploying an application via YAML blueprint consisting of the following layers.
+Below is the annotated blueprint. [Download the blueprint](example_yaml/mycluster.yaml) into a 
+text file, `mycluster.yaml`, in your workspace. *Before* you create an application with it, 
+review and/or change the the location where the application will be deployed.
 
-- A dynamically scalable Web App Cluster
-- A MySQL DB
-
-
-Copy the blueprint below into a text file, "mycluster.yaml", in your workspace, but *before* you create an application 
-with it, again modify the YAML to specify the location where the application will be deployed.  
-You will need at least five machines for this example, one for the DB, one for the Nginx controller, and three for the tomcats 
-(but you can reduce this by changing the "maxPoolSize" below.
-
-{% highlight yaml %}
-name: cluster
-
-location:
-  jclouds:aws-ec2:
-    identity: ABCDEFGHIJKLMNOPQRST
-    credential: s3cr3tsq1rr3ls3cr3tsq1rr3ls3cr3tsq1rr3l
+You will need four machines for this example: one for the load-balancer (nginx), and three for the 
+Tomcat cluster (but you can reduce this by changing the `maxPoolSize` below).
 
 
-services:
-- serviceType: brooklyn.entity.webapp.ControlledDynamicWebAppCluster
-  name: webcluster
+<!-- WARNING: if modifying either mycluster.yaml or the yaml below, be sure to keep them both in-sync -->
+
+<div class="jumobotron annotated_blueprint" markdown="1">
+  <div class="code_scroller">
+    <div class="initial_notice"><div><div>
+      Hover over an element to learn more
+      <div class="ann_light">This message will go away in <span id="countdown">3s</span></div>
+      <div class="ann_play fa fa-play-circle-o"></div>
+    </div></div></div>
+    <div class="code_viewer">
+  
+<div class="block">
+      <div class="annotations_wrapper1"><div class="annotations_wrapper2"><div class="annotations">
+        <div class="short">
+          Describe your application
+        </div>
+        <div class="long"><p>
+            Start by giving it a name, optionally adding a version and other metadata.
+        </p></div>
+      </div><div class="connector"><div>&nbsp;</div></div></div></div>
+<div><span class="ann_highlight">name: Tomcat Cluster</span>
+</div></div>
+
+<div class="block">
+      <div class="annotations_wrapper1"><div class="annotations_wrapper2"><div class="annotations">
+        <div class="short">
+          Define the target location
+        </div>
+        <div class="long"><p>
+          Blueprints are designed for portability.
+          Pick from dozens of clouds in hundreds of datacenters. 
+          Or machines with fixed IP addresses, localhost, 
+          Docker on <a href="http://clocker.io">Clocker</a>, etc.
+        </p><p>
+          Here we target pre-existing Vagrant VMs.
+        </p></div>
+      </div><div class="connector"><div>&nbsp;</div></div></div></div>
+<div><span class="ann_highlight">location:</span>
+  byon:
+    user: vagrant
+    password: vagrant
+    hosts:
+      - 10.10.10.101
+      - 10.10.10.102
+      - 10.10.10.103
+      - 10.10.10.104
+</div></div>
+
+<div class="block">
+      <div class="annotations_wrapper1"><div class="annotations_wrapper2"><div class="annotations">
+        <div class="short">
+          Define a cluster
+        </div>
+        <div class="long"><p>
+            Choose your cluster type.
+          </p><p>
+            Customize with config keys, such as the initial size. Define the members of the cluster.
+        </p></div>
+      </div><div class="connector"><div>&nbsp;</div></div></div></div>
+<div>services:
+<span class="ann_highlight">- type: org.apache.brooklyn.entity.group.DynamicCluster</span>
+  name: Cluster
+  id: cluster
   brooklyn.config:
-    wars.root: http://search.maven.org/remotecontent?filepath=io/brooklyn/example/brooklyn-example-hello-world-sql-webapp/0.6.0-M2/brooklyn-example-hello-world-sql-webapp-0.6.0-M2.war
-    http.port: 9280+
-    proxy.http.port: 9210+
-    java.sysprops: 
-      brooklyn.example.db.url: $brooklyn:formatString("jdbc:%s%s?user=%s\\&password=%s",
-         component("db").attributeWhenReady("datastore.url"), "visitors", "brooklyn", "br00k11n")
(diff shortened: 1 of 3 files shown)


guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
 - { section: "Managing with Policies", title_in_menu: "Policies"  } 
 ---
 
+## A Clustered Example
 
+We'll now look at a more complex example that better shows the capabilities of Brooklyn, including
+inlife management of a clustered application.
```
drigodwin: I would hyphenate in-life or use different wording such as "including management of a running clustered application"

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
+
+Below is the annotated blueprint.
+
+Copy the blueprint into a text file, `mycluster.yaml`, in your workspace, but *before* you create 
+an application with it, review and/or change the the location where the application will be
```
drigodwin: I would put the break in later "Below is an annotated blueprint, copy this blueprint into a text file, `mycluster.yaml`, in your workspace. Before..."

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
-      
-- serviceType: brooklyn.entity.database.mysql.MySqlNode
-  id: db
-  name: mysql
-  brooklyn.config:
-    creationScriptUrl: https://bit.ly/brooklyn-visitors-creation-script
+Deploy the app:
+
+{% highlight bash %}
+br deploy mycluster.yaml
```
drigodwin: if you put in a $ at the start of the bash lines it'll be consistent with the rest of the docs and is ignored by the copy function

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
+br application "Tomcat Cluster" entity "Cluster" entity
 {% endhighlight %}
+ 
+<pre>
+ Id         Name            Type   
+ dYfUvLIw   quarantine      org.apache.brooklyn.entity.group.QuarantineGroup   
+ tOpMeYYr   Tomcat Server   org.apache.brooklyn.entity.webapp.tomcat.TomcatServer   
+</pre>
+
+The "quarantine" entity is used when Tomcat servers fail to start correctly - then entity is by
```
drigodwin: then entity -> this entity

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
+default added to the quarantine group, where it can later investigated. This can be disabled using
+the configuration `dynamiccluster.quarantineFailedEntities: false`.
+
 
-the syntax ```$brooklyn:formatString(...)``` is an example of the Brooklyn DSL (Domain Specific Language) which 
-allows expressions referring to Brooklyn's management information to be embedded in blueprints.  The line above also illustrates the use of Brooklyn's ```component(...)``` and ```attributeWhenReady(...)``` to get an identified component from a deployment, and to wait until the component is fully deployed before reading one of its sensors ("datastore.url" in this case). 
+## Tomcat auto-repair
 
-## Managing with Policies
+Each Tomcat entity has a `ServiceFailureDetector` enricher and a `ServiceRestarter` policy.
```
drigodwin: Have we talked about enrichers anywhere before this?

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
 
+TODO Why doesn't the restart() show in the activity view?!
+-->
+
+After the given `stabilizationDelay`, the Tomcat server will be automatically restarted.
+
+
+## Cluster auto-replace
+
+The cluster has a `ServiceReplacer` policy. This attaches to a DynamicCluster and replaces a
```
drigodwin: I would highlight DynamicCluster in the same way throughout the document

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
+TODO Why doesn't the restart() show in the activity view?!
+-->
+
+After the given `stabilizationDelay`, the Tomcat server will be automatically restarted.
+
+
+## Cluster auto-replace
+
+The cluster has a `ServiceReplacer` policy. This attaches to a DynamicCluster and replaces a 
+failed member; if this fails, it sets the Cluster state to on-fire.
```
drigodwin: I don't think 'on-fire' has been explained anywhere before - perhaps the section above about failure is a good place to do this?

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
 
-The loadbalancer's `Controller targets tracker` policy ensures that the loadbalancer is updated as the cluster size changes.
+To simulate a terminal the failure of a member, repeatedly kill the process (using the command
```
drigodwin: simulate a terminal the failure -> simulate a terminal failure

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
-### Under the Covers
+## Auto-scaling
 
-The `AutoScalerPolicy` here is configured to respond to the sensor
-reporting requests per second per node, invoking the default ```resize``` effector.
-By updating on the policy, you can configure it to respond to a much lower threshhold
-or set long stabilization delays (the period before it scales out or back).
+The `AutoScalerPolicy` attaches to a `Resizable` entity (in this case the `DynamicCluster`) and 
+dynamically adjusts its size in response to either keeping a metric within a given range, or in 
+response to Pool Cold and Pool Hot events. It adds/removes members (e.g. Tomcat instances)
```
drigodwin: What are pool hot or cold events? Could there be a link to a definition?

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
+very low at just 3. This makes it simple to trigger auto-scaling. The `resizeUpStabilizationDelay` 
+of 2 seconds means the load must be sustained for at least that length of time. The  
+`resizeDownStabilizationDelay` of 1 minute means there must be low load for a full one minute 
+before it will scale back down.
 
-However it is possible to manually suspend the policy, by 
+Find the URL of the load-balancer - see the `NginxController` sensor named `main.uri`.
+
+To generate load, you can use your web-browser by repeatedly refreshing that page. Alternatively,
+use a script like that below, or use a load generator like jmeter:
```
drigodwin: Alternatively, use a script like that below -> Alternatively, use a script such as the one shown below (changing URL for the URL of your load-balancer)

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
+Below is the annotated blueprint.
+
+Copy the blueprint into a text file, `mycluster.yaml`, in your workspace, but *before* you create 
+an application with it, review and/or change the the location where the application will be 
+deployed.
+
+You will need four machines for this example: one for the load-balancer (nginx), and three for the 
+Tomcat cluster (but you can reduce this by changing the "maxPoolSize" below).
+
+<!-- TODO How to make this example's copy-paste work properly?! -->
```
drigodwin: Perhaps you could just have the mycluster.yaml to download from a link?

guide/start/policies.md
```diff
@@ -9,169 +9,423 @@ children:
...
-For example
-{% highlight yaml %}
-$ br app cluster ent webcluster policy
-Id         Name                                                      State   
-mMZngBnb   org.apache.brooklyn.policy.autoscaling.AutoScalerPolicy   RUNNING   
+Try killing the Tomcat process for one of the members in the cluster. The command below will kill
+Tomcat on the vagrant VM named "byon1":
+
+{% highlight bash %}
+vagrant ssh byon1 --command 'ps aux | grep -i tomcat |  grep -v grep | awk '\''{print $2}'\'' | xargs kill -9'
```
drigodwin: So on mine tomcat was running on byon2 not byon1

drigodwin:
That's really good, I think it's useful and concise explanation - I've made a few specific points.

guide/start/policies.md
aledsage: I'll take it out of the getting started guide - a separate how-to (not yet written!) about the auto-scaling policy could talk about more advanced use-cases.

FYI the hot/cold events are a way that an external system (or external enricher) can trigger auto-scaling. Instead of having some metric kept in a given range, something else can figure out what needs to be done and then emit these events to tell the auto-scaler that the cluster is too hot (i.e. over-utilised, so too small to handle the current workload) or too cold (i.e. under-utilised). This is only used by a small number of (old) entities.

aledsage:
Thanks @drigodwin - great comments. I've incorporated all of those, squashed the commits and rebased.

guide/start/policies.md
```diff
@@ -9,169 +9,438 @@ children:
...
+
+## Tomcat auto-repair
 
-## Managing with Policies
+Each Tomcat entity has a `ServiceFailureDetector` enricher and a `ServiceRestarter` policy. 
 
+An *enricher* generates new events or sensor values (i.e. metrics), e.g. by aggregating existing 
+metrics; it enriches the sensor data available for that entity. A *policy* performs 
+runtime management: it performs actions on an entity (or entities), for example to keep the system 
+in a healthy state.
```
drigodwin: How about

An *enricher* generates new events or sensor values (metrics) which it uses to populate the sensor data available for a specific entity. A *policy* coordinates the runtime management of entities by carrying out actions initiated by specific triggers. This may be to keep the system in a healthy state, or autoscale a cluster for example.

guide/start/policies.md
```diff
@@ -9,169 +9,438 @@ children:
...
+and similarly an "Entity Recovered" event when recovered. The configuration option 
+`serviceOnFireStabilizationDelay` will suppress reporting of failure until the entity is detected 
+as failed for the given duration. This is very useful so as not to over-react to temporary failures.
+
+The `ServiceRestarter` policy attaches to an entity in order to restart the service on failure. If  
+there are subsequent failures within a configurable time interval, or if the restart fails, the  
+service entity is marked as failed and no futher restarts are attempted.
+
+Try killing the Tomcat process for one of the members in the cluster. The command below will kill
+Tomcat on the vagrant VM named "byon1":
```
drigodwin: on the vagrant VMs named byon1 to byon4

aledsage:
@drigodwin thanks again - incorporated those last changes and have merged.

PR #6, opened 2016-02-02
Remove incubating

aledsage:


ahgittin:
looks good.  haven't tested it all but we'll do that on an ongoing basis.

PR #8, opened 2016-02-03
get the summary script from our codebase, not from a gist

ahgittin:
NB: depends on https://github.com/apache/brooklyn-dist/pull/2


ahgittin:
being applied as part of refactoring suggested by #1 review; closing (but you'll see the changes soon)

PR #9, opened 2016-02-03
Documentation for initial support for SaltStack.

geomacy:
See https://issues.apache.org/jira/browse/BROOKLYN-217.

    This commit merges code from incubator-brooklyn, following the move to the new repos
    after graduation. As such, it unfortunately loses the commit history, but the code
    is based on the work done by Hadrian Zbarcea.  See the Jira link above for history
    and discussion.

### DIFF
--- a/guide/yaml/index.md
+++ b/guide/yaml/index.md
@@ -12,6 +12,7 @@ children:
 - custom-entities.md
 - winrm/
 - chef/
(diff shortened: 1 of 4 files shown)


ahgittin:
docs for https://github.com/apache/brooklyn-library/pull/1

PR #10, opened 2016-02-05
document location inheritance better, including recent fix

ahgittin:
documents the clarification implemented in https://github.com/apache/brooklyn-server/pull/7


ahgittin:
@neykov i see you merged https://github.com/apache/brooklyn-server/pull/7 -- can you take a gander at this (small) one?

PR #11, opened 2016-02-08
Fix get the code

ahgittin:
Address @rdowner 's comments on #1


website/developers/code/git-more.md
```diff
@@ -0,0 +1,297 @@
...
+If you're contributing or working a lot on a feature, 
+you'll probably want your own forks and a slightly different git remote setup.
+
+You can create forks of each Brooklyn repo [in the GitHub UI](https://github.com/apache?query=brooklyn) 
+or, if you have the command-line tool `hub` ([described here](https://github.com/github/hub), or `sudo npm install -g hub`),
+by running this command:
+
+{% highlight bash %}
+hub fork; git submodule foreach 'hub fork'
+{% endhighlight %}
```
richardcloudsoft: Something I have just discovered worth noting! If you already have forks, this command still works - your fork is left alone but it still sets up the new remote.

website/developers/code/git-more.md
```diff
@@ -0,0 +1,297 @@
...
+
+{% highlight bash %}
+git remote rename origin upstream; git submodule foreach 'git remote rename origin upstream'
+{% endhighlight %}
+
+You'll now likely want to add the remote `origin` for your fork:
+
+{% highlight bash %}
+if [ -z "$GITHUB_ID" ] ; then echo -n "Enter your GitHub ID id: " ; read GITHUB_ID ; fi
+git remote add origin git@github.com:${GITHUB_ID}/apache
```
richardcloudsoft: Why not simply do `git remote rename ahgittin origin`?

If you prefer to keep this, then I guess it should be `/brooklyn` on the end instead of `/apache`.

website/developers/code/git-more.md
```diff
@@ -0,0 +1,297 @@
...
+{% endhighlight %}
+
+And also optionally, to set up the official Apache repo as a remote ---
+useful if GitHub is slow to update (and required if you're a committer):
+ 
+{% highlight bash %}
+git remote add apache-git https://git-wip-us.apache.org/repos/asf/brooklyn
+git submodule foreach 'git remote add apache-git https://git-wip-us.apache.org/repos/asf/${name}'
+{% endhighlight %}
+
```
richardcloudsoft: I'd still prefer that branches for PRs and the canonical Apache repository is not in this "getting started" workflow (even if it is marked as optional). I'm happy for it to be later on in this doc though - but under its own header further down, I just don't think it should be in the first section.

website/developers/code/git-more.md
```diff
@@ -0,0 +1,297 @@
...
+You'll now likely want to add the remote `origin` for your fork:
+
+{% highlight bash %}
+if [ -z "$GITHUB_ID" ] ; then echo -n "Enter your GitHub ID id: " ; read GITHUB_ID ; fi
+git remote add origin git@github.com:${GITHUB_ID}/apache
+git submodule foreach 'git remote add origin git@github.com:${GITHUB_ID}/${name}'
+{% endhighlight %}
+
+And if you created the fork in the GitHub UI, you may want to create a remote named by your
+GitHub ID as well (if you used `hub` it will have done it for you):
```
richardcloudsoft: So you are aiming to end up with a remote called `origin` and a remote called `ahgittin` both pointing at the same repo? I think that's unnecessary complexity. (Not that it is especially complex per se - but I strongly subscribe to [Antoine de Saint-Exupéry's principle](https://www.goodreads.com/quotes/19905-perfection-is-achieved-not-when-there-is-nothing-more-to)!)

Keep two paths - one using `hub fork` and one not, and make the end state the same for both. Amongst git tutorials common practice is `origin` and `upstream` so let's use those. The power users like you and me will know what to do if we want things different :smile:

website/developers/how-to-contribute.md
```diff
@@ -6,7 +6,8 @@ children:
 - { section: Create an Issue in Jira, title: Jira }
 - { section: Contributing using GitHub, title: GitHub }
 - { section: Reviews }
-- { section: Handy Places }
+- { link: code/, title: Get the Code }
+- { link: links.html, title: Handy Places }
```
richardcloudsoft: These two links are now duplicated in the menu - they can be removed from here.
screenshot 2016-02-09 13 07 36

website/developers/how-to-contribute.md
```diff
@@ -106,8 +108,10 @@ $ git clone https://git-wip-us.apache.org/repos/asf/brooklyn-server.git
...
 attributed as the author of the patch when a committer merges it.
+The review process will be as with pull requests, except for comments only appearing on the Jira issue.
 
 
-## Handy Places
+### Handy Places
 
-Development bookmarks are available [here](links.html).
+If you've not done so, you'll probably want to start by [getting the code](code/).
+Once you've done that, you'll find [handy development bookmarks here](links.html).
```
richardcloudsoft: I think this section can be dropped. It duplicates other information in the side menu, and is such a small section it probably won't be noticed by many people.

website/developers/links.md
```diff
@@ -1,20 +1,18 @@
...
 
-* **Maven repositories** are:
+* **Maven repositories** are at:
+
   * [Apache releases]({{ apache_releases_repo_groupid_url }})
   * [Apache snapshots]({{ apache_snapshots_repo_groupid_url }})
- 
-* **CI server** is a jenkins server at: [https://builds.apache.org/job/brooklyn-master-build/](https://builds.apache.org/job/brooklyn-master-build/)
+
+* **CI** is done by Jenkins at [https://builds.apache.org/job/brooklyn-master-build/](https://builds.apache.org/job/brooklyn-master-build/)
```
richardcloudsoft: I think this file can be dropped completely. GitHub and JIRA links are duplicated information as they are already in the menu; Jenkins isn't but it should be. There shouldn't be any need to refer to the Apache releases Maven repo as it pushes to Maven Central. That just leaves the Apache snapshot repo, and I expect we can find a better place to document that.

richardcloudsoft:
Thanks @ahgittin, this improves usability(readability!) a lot. There's still some issues remaining, but IMO it's worth publishing these changes ASAP so I'm going to merge now and regenerate the main website.

richardcloudsoft:
Some things to add after my experiences with submodules...

I think that this command is necessary after the first `git submodule update`:

```bash
git submodule foreach "git checkout master"
```

Apache committers may benefit from these commands - it overrides the `pushDefault` (which is origin, or my personal fork) for the master branch to be the canonical Apache repo, so a simple `git push` on master will go to the canonical repo. (On the other hand, there's a good argument for *not* doing this, and requiring the apache-git remote name to be explicitly specified, to avoid accidentally pushing to the wrong place...)

```bash
git config branch.master.pushRemote apache-git
git submodule foreach "git config branch.master.pushRemote apache-git"
```

ahgittin:
@richardcloudsoft I've applied and pushed the essential changes in https://github.com/apache/brooklyn-docs/commit/fce745edd47399d6b4ec550d81cf37997f0eadf1 .

Regarding `apache-git` as the default push target -- I really do **not** want that.  I'm irritated as it is that the new version of `git` caches my password!  I like the extra checks.  Though now that force push is allowed it's not quite so grave.

As for the other comments I could go either way.  Probably slightly prefer it as it is, longwinded but thorough, but happy if someone cleans it up further.  (OTOH removing the script to make it like this is was a great suggestion!  Thanks.)

richardcloudsoft:
Ok that's great, I think we're done for now. Thanks for taking the time to make the changes - I think it's made the first (developer) contact much more straightforward!

PR #12, opened 2016-02-10
use local version of zeroclipboard movie

ahgittin:
fixes issues in some recent browser (eg latest chrome on latest os x)
    where the flash "movie" won't load unless it is local.
    the error it gives is a generic "uncaught exception" in
    ZeroClipboard activate -> flashState.bridge.setSize.
    curiously it still works fine in some browsers (firefox) and
    older chrome was also fine but I'm guessing
    security modes are being tightened in browsers.

    none of the other recommended settings for fixing this seem to work:
    * trustedDomains
    * trustedOrigin
    * allowScriptAccess
    * and even the flash config in the browser

    it's just a small movie however so should be low impact (smaller than most of our images).
    put in img path to simplify access.

    also, zeroclipboard is updated from 1.3.1 to 1.3.5 (latest stable 1.x branch).
    (2.x branch looks like it might be a bigger api change although some fixes in this area are reported.)


ahgittin:
FYI almost identical change made (just now) to docs at https://github.com/apache/brooklyn-docs/pull/12

ahgittin:
oops previous comment should have been https://github.com/apache/brooklyn-ui/pull/3#issuecomment-183264340 

(this PR is of course identical to itself!)

closing for now while we resolve the binaries-in-source issue as discussed at the issue above

PR #13, opened 2016-02-12
Initial Ansible integration documents.

geomacy:
These are the document updates for the Ansible integration work, they describe in broad detail the way Brooklyn can now integrate with Ansible, and what configuration  options there are for ansible within blueprints.


ygy:
Looks very good

PR #16, opened 2016-02-18
use local version of zeroclipboard movie

ahgittin:
fixes issues in some recent browser (eg latest chrome on latest os x)
    where the flash "movie" won't load unless it is local.
    the error it gives is a generic "uncaught exception" in
    ZeroClipboard activate -> flashState.bridge.setSize.
    curiously it still works fine in some browsers (firefox) and
    older chrome was also fine but I'm guessing
    security modes are being tightened in browsers.

    none of the other recommended settings for fixing this seem to work:
    * trustedDomains
    * trustedOrigin
    * allowScriptAccess
    * and even the flash config in the browser

    it's just a small movie however so should be low impact (smaller than most of our images).
    put in img path to simplify access.

    also, zeroclipboard is updated from 1.3.1 to 1.3.5 (latest stable 1.x branch).
    (2.x branch looks like it might be a bigger api change although some fixes in this area are reported.)

    note the SWF binary is downloaded on first run of _build/build.sh,
    similar to as we do for the brooklyn-ui project, and also with a README in that dir


ahgittin:
update of #12 addressing @rdowner 's points (which may not be necessary here as this is not in the source build, but seems good practise).

nearly identical to https://github.com/apache/brooklyn-ui/pull/8 , just for the docs, and relies on `curl` in the `_build/build.sh` script.

ahgittin:
merging as this is a fix of a previous PR

PR #18, opened 2016-02-23
add mailing list search box to website

johnmccabe:
- uses http://markmail.org/
- spotted the idea on http://couchdb.apache.org


ahgittin:
love it

PR #21, opened 2016-03-01
WinRM documentation - run as Administrator

bostko:


aledsage:
Thanks @bostko - merging this now, and then I'll make some edits to the wording. I presume this new text is very relevant to the section "How and Why to re-authenticate within a powershell script", so I want to make sure we capture it there as well.

PR #23, opened 2016-03-03
Printable PDF Generation

drigodwin:
Produces a second, more printer friendly version of the PDF documentation. Uses an additional style sheet to remove colours and formatting unsuitable for printing.

### DIFF
--- a/_build/build.sh
+++ b/_build/build.sh
@@ -191,12 +191,12 @@ function build_pdf() {
   PDF_STARTED_LOG="_pdf/pdf_gen_started.log"
   
   if [ "$SKIP_PDF_MANUAL" != "true" ]; then
-    _build/buildPDF.sh "_site/zoneMergeManual.html" "_pdf/UserManual.pdf" 2>&1 | tee $PDF_MANUAL_LOG
-    _build/buildPDF.sh "_site/zoneMergeManual.html" "_pdf/UserManual_printable.pdf" "./style/css/printable.css" 2>&1 | tee $PDF_MANUAL_LOG
+    _build/build-pdf.sh "_site/zoneMergeManual.html" "_pdf/UserManual.pdf" 2>&1 | tee $PDF_MANUAL_LOG
+    _build/build-pdf.sh "_site/zoneMergeManual.html" "_pdf/UserManual_printable.pdf" "./style/css/printable.css" 2>&1 | tee $PDF_MANUAL_LOG
   fi
   if [ "$SKIP_PDF_STARTED" != "true" ]; then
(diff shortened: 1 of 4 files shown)


_build/build.sh
```diff
@@ -192,9 +192,11 @@ function build_pdf() {
   
   if [ "$SKIP_PDF_MANUAL" != "true" ]; then
     _build/buildPDF.sh "_site/zoneMergeManual.html" "_pdf/UserManual.pdf" 2>&1 | tee $PDF_MANUAL_LOG
+    _build/buildPDF.sh "_site/zoneMergeManual.html" "_pdf/UserManual_printable.pdf" "./style/css/printable.css" 2>&1 | tee $PDF_MANUAL_LOG
```
ahgittin: not related to this PR but is there a reason this isn't `build-pdf.sh` consistent with the other commands in the dir?

_build/buildPDF.sh
```diff
@@ -18,7 +18,11 @@ if ! command -v wkhtmltopdf >/dev/null 2>&1; then
 fi
 	
 # run the PDF build with wkhtmltopdf
-wkhtmltopdf --page-size Letter --margin-top 0.75in --margin-right 0.75in --margin-bottom 0.75in --margin-left 0.75in --encoding UTF-8 $1 $2
+if [ -z "$3" ] ; then
+	wkhtmltopdf --page-size Letter --margin-top 0.75in --margin-right 0.75in --margin-bottom 0.75in --margin-left 0.75in --encoding UTF-8 $1 $2
+else
+	wkhtmltopdf --user-style-sheet $3 --page-size Letter --margin-top 0.75in --margin-right 0.75in --margin-bottom 0.75in --margin-left 0.75in --encoding UTF-8 $1 $2
```
ahgittin: could revert from `if/then` to a single line using `${parameter:+word}`

    wkhtmltopdf ${3:+--user-style-sheet $3} --page-size Letter ...

(i can never remember this bash magic, i always do `man bash` then search for `##` to find parameter expansion...)

style/css/printable.css
```diff
@@ -0,0 +1,9 @@
+/*
+ * This is a CSS included in the printable version of docs, it's used to mask out large images etc
+ */
```
ahgittin: not sure this does anything to large images yet but makes sense that would be the intention (no need to change if it is)

ahgittin:
very minor, then good to merge

drigodwin:
Thankyou for the comments @ahgittin !

aledsage:
+1; LGTM - merging now.

PR #25, opened 2016-03-09
Cli doc updates

geomacy:


guide/ops/cli/index.md
```diff
@@ -6,7 +6,25 @@ children:
...
 information on starting on a Brooklyn Server, refer to [Server CLI Reference](../server-cli-reference.html).
+
+## Obtaining the CLI tool
+
+The CLI tool, `br`,  is included in the Apache Brooklyn distribution, in bin/brooklyn-client-cli-0.9.0-SNAPSHOT.
+The tool is an executable binary (written in Go), and is distributed in a number of builds for a variety of platforms
+and architectures, each build in its own subdirectory:
+
+    darwin.386    freebsd.386    linux.386    netbsd.386     openbsd.386    windows.386
+    darwin.amd64  freebsd.amd64  linux.amd64  netbsd.amd64   openbsd.amd64  windows.amd64
```
ahgittin: would it be (a) easy and (b) sensible just to do:

* darwin.amd64 (Mac)
* linux.{386,amd64}
* windows.{386,amd64}

i don't think we've ever discussed tested or had a request for the others.  might improve build time by 1m+ ?

robertgmoss: BTW @ahgittin @geomacy I missed this before: I'm pretty sure the windows version won't work properly since it relies on putting a file in the $HOME directory for login.

neykov: re windows - the executable doesn't have an `.exe` extension.

guide/ops/cli/index.md
```diff
@@ -6,7 +6,25 @@ children:
...
 
-{% include list-children.html %}
 
-**NOTE:** These documents are for using the Brooklyn Client CLI to access a running Brooklyn Server.  For 
+**NOTE:** These documents are for using the Brooklyn Client CLI tool to access a running Brooklyn Server.  For
 information on starting on a Brooklyn Server, refer to [Server CLI Reference](../server-cli-reference.html).
+
+## Obtaining the CLI tool
+
+The CLI tool, `br`,  is included in the Apache Brooklyn distribution, in bin/brooklyn-client-cli-0.9.0-SNAPSHOT.
```
ahgittin: we need `<!-- BROOKLYN_VERSION_BELOW -->` on the previous line

guide/ops/cli/index.md
geomacy: I'll amend the build to reflect the above, and also exclude the windows until it's tested and working.

ahgittin: i think supporting windows CLI is a priority.  suggest fixing it rather than changing the docs.

will you also exclude the 6 bsd items and the 32-bit darwin item?  assuming yes i'll update the other PR's and merge.

geomacy: Yes I'll do that.  Will test the windows version and fix.

geomacy: @robertgmoss I will have a look at how to get round that $HOME problem, maybe use %USERPROFILE% instead(?)

ahgittin:
i will tweak dist and docs so that:

* it puts binaries in `bin/brooklyn-client-cli/` (unversioned dir)
* we only list the binaries we care about

then merge

will then do a follow-up which adjusts how the client build and files are excluded, as discussed at https://github.com/apache/brooklyn-dist/pull/14

guide/start/running.md
```diff
@@ -133,9 +134,9 @@ INFO  Started Brooklyn console at http://127.0.0.1:8081/, running classpath://br
 
 By default it can be accessed by opening [127.0.0.1:8081](http://127.0.0.1:8081){:target="_blank"} in your web browser. 
 
-The rest of this getting started guide uses the Apache Brooklyn command line interface (CLI). To use this, download and install the tool as described on the [CLI GitHub page](https://github.com/brooklyncentral/brooklyn-cli){:target="_blank"}.
+The rest of this getting started guide uses the Apache Brooklyn command line interface (CLI) tool, `br`. 
+This tool is distributed along with Apache Brooklyn. See details on the [Client CLI Reference](../../ops/cli/) page.
```
drigodwin: I think URL is incorrect

guide/ops/cli/index.md
geomacy: Noting here for the sake of completeness, that @murdoaird points out the Windows build works - you just have to rename the binary as br.exe and define the HOME environment variable.

geomacy: Updated build to just create the 32 bit versions of linux/windows, and the (64 bit) Mac:

https://github.com/apache/brooklyn-client/pull/14

(Just 32 bit for linux/windows as it keeps releases smaller, the 32 bit version should be fine for each.)

PR #27, opened 2016-03-10
Update maven-build.md with RPM notes.

alrick87:
Add notes about building RPM packages.


aledsage:
LGTM; I'll add a few more pointers that I needed as well after merging.

For the record, this is needed after we merge https://github.com/apache/brooklyn-dist/pull/13

PR #28, opened 2016-03-11
Clarified catalog default

drigodwin:


aledsage:
For things like this, I'd squash the two commits together and then `push -f` to your personal branch. In the git history, the message "typo" is pretty meaningless when not in looking at the specific PR.

Merging now anyway.

PR #31, opened 2016-03-18
Describe SoftLayerSameVlanLocationCustomizer configuration

grkvlt:
Documentation for the changes made in https://github.com/apache/brooklyn-server/pull/53

### DIFF
--- a/guide/ops/locations/more-clouds.md
+++ b/guide/ops/locations/more-clouds.md
@@ -172,4 +172,103 @@ For example, for dev/demo purposes an "everything" network could be created that
 || Source IP Ranges            || 0.0.0.0/0                   |
 || Allowed protocols and ports || tcp:0-65535 and udp:0-65535 |
 
+## IBM SoftLayer
 
+### VLAN Selection
+
+SoftLayer may provision VMs in different VLANs, even within the same region.
+Some applications require VMs to be on the *same* internal subnet; blueprints
+for these can specify this behaviour in SoftLayer in one of two ways.
+
+The VLAN ID can be set explicitly using the fields
+`primaryNetworkComponentNetworkVlanId` and
+`primaryBackendNetworkComponentNetworkVlanId` of `SoftLayerTemplateOptions`
+when specifying the location being used in the blueprint, as follows:
+
+```YAML
+location:
+  jclouds:softlayer:
+    region: ams01
+    templateOptions:
+      # Enter your preferred network IDs
+      primaryNetworkComponentNetworkVlanId: 1153481
+      primaryBackendNetworkComponentNetworkVlanId: 1153483
+```
+
+This method requires that a VM already exist and you look up the IDs of its
+VLANs, for example in the SoftLayer console UI, and that subsequently at least
+one VM in that VLAN is kept around.  If all VMs on a VLAN are destroyed
+SoftLayer may destroy the VLAN.  Creating VLANs directly and then specifying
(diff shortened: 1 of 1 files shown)


guide/ops/locations/more-clouds.md
```diff
@@ -172,4 +172,86 @@ For example, for dev/demo purposes an "everything" network could be created that
 || Source IP Ranges            || 0.0.0.0/0                   |
 || Allowed protocols and ports || tcp:0-65535 and udp:0-65535 |
 
-
+## IBM SoftLayer
+
+### VLAN Selection
+
+For some applications it is important that All VMs use the same VLAN. On
```
ahgittin:
```
SoftLayer may provision VMs in different VLANs, even within the same region.
Some applications require VMs to be on the *same* internal subnet;
blueprints for these can specify this behaviour in SoftLayer in one of two ways.

The VLAN ID can be set explicitly using the fields ...
```

guide/ops/locations/more-clouds.md
```diff
@@ -172,4 +172,86 @@ For example, for dev/demo purposes an "everything" network could be created that
...
+  jclouds:softlayer:
+    region: ams01
+    templateOptions:
+      # Enter your preferred network IDs
+      primaryNetworkComponentNetworkVlanId: 1153481
+      primaryBackendNetworkComponentNetworkVlanId: 1153483
+```
+
+This will only work if you already have a VM configured using the same
+VLAN, and can copy the IDs from the SoftLayer console UI.
```
ahgittin:
```
This method requires that a VM already exist and you look up the IDs of its VLANs, for example in the SoftLayer console UI, and that subsequently at least one VM in that VLAN is kept around.  If all VMs on a VLAN are destroyed SoftLayer may destroy the VLAN.  Creating VLANs directly and then specifying them as IDs here may not work.
```

guide/ops/locations/more-clouds.md
```diff
@@ -172,4 +172,86 @@ For example, for dev/demo purposes an "everything" network could be created that
...
+
+If you are deploying a new application, then you can customise the
+location with ` `JcloudsLocationCustomizer` class, and we provide an
+implementation that ensures all machines in a particular scope will
+use the same VLAN. By default, the scope used is the application ID
+for the blueprint being deployed, but this can be changed if you need
+multiple applications to share the same VLAN.
```
ahgittin:
```
The second method tells Brooklyn to discover VLAN information automatically:
it will provision one VM first, and use the VLAN information from it when provisioning subsequent machines.
This ensures that all VMs are on the same subnet without requiring any manual VLAN referencing,
making it very easy for end-users.
```

(also note extra backtick in line 199)

guide/ops/locations/more-clouds.md
```diff
@@ -172,4 +172,86 @@ For example, for dev/demo purposes an "everything" network could be created that
...
+
+Use the `SoftLayerSameVlanLocationCustomizer` in your blueprint by
```
ahgittin:
```
To use this method, we tell brooklyn to use `SoftLayerSameVlanLocationCustomizer` as a location customizer.  
This can be done on a location as follows:
```

(put location first, then)

```
It is also possible with many blueprints to specify this as one of the `provisioning.properties` on an *application*:
```

(app blueprint illustration)

ahgittin:
great content -- quite a few minor tweaks

then good to merge once the change in server is merged

PR #32, opened 2016-03-22
add -P prefix to invoke params in br reference/example

johnmccabe:
prefix was missing from docs


aledsage:
Thanks @johnmccabe - merging.

PR #33, opened 2016-03-24
Update running.md

alrick87:


aledsage:
Thanks - merging. @murdoaird can build on this, to give more info to those unfamiliar with RPMs and the like.

PR #34, opened 2016-03-24
Minor wording changes around option of installing binaries in own env

murdoaird:
Calling it "local install" was confusing - people didn't realise you could install the binaries anywhere.


aledsage:
LGTM; merging.

PR #36, opened 2016-03-24
Details about using CredSSP in Windows

bostko:
@aledsage can you review.

### DIFF
--- a/guide/yaml/winrm/index.md
+++ b/guide/yaml/winrm/index.md
@@ -289,11 +289,25 @@ When a script is run over WinRM, the credentials under which the script are run 
 solution is to obtain a new set of credentials within the script and use those credentials to 
 required commands.
 
-Certain Windows registry keys must be reconfigured in order to support re-authentication. For 
-clouds that support an init script, Brooklyn can take care of this at instance boot time, as part 
-of the setup script. For clouds where an init script is not (currently) supported, such as Azure, 
-it is assumed that the VM is already correctly configured. Please ensure that Brooklyn's changes 
-are compatible with your organisation's security policy.
+The WinRM client uses Negotiate+NTLM to authenticate against the machine.
+This mechanism applies certain restrictions to executing commands on the windows host.
+
+For this reason you should enable CredSSP on the windows host which grants all privileges available to the user.
+ https://technet.microsoft.com/en-us/library/hh849719.aspx#sectionSection4
+
+To use `Invoke-Command -Authentication CredSSP` the Windows Machine has to have:
+- Up and running WinRM over http. Notice that we support winrm over https but for Invoke-Command to work it needs up and running winrm over http.
+  Apache Brooklyn can winrm over https but if the install script has inside it `Invoke-Command -Authentication CredSSP` then winrm over http has to be enabled as well.
+- Added trusted host entries which will use Invoke-Command
+- Allowed CredSSP
+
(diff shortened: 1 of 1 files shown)


guide/yaml/winrm/index.md
```diff
@@ -289,11 +289,24 @@ When a script is run over WinRM, the credentials under which the script are run
...
-it is assumed that the VM is already correctly configured. Please ensure that Brooklyn's changes 
-are compatible with your organisation's security policy.
+The WinRM client uses Negotiate+NTLM to authenticate against the machine.
+This mechanism applies certain restrictions to executing commands on the windows host.
+
+For this reason you should enable CredSSP on the windows host which grants all privileges available to the user.
+ https://technet.microsoft.com/en-us/library/hh849719.aspx#sectionSection4
+
+To use `Invoke-Command -Authentication CredSSP` the Windows Machine has to have:
+- Up and running WinRM over http
```
aledsage: or https?

bostko: http as well

aledsage: So can you change the wording so that it does not imply that it must be http?

bostko: I am saying that is must have http enabled.

cloudsoftreleaser: @bostko I thought we were using https with winrm4j now. Are we using http just for this? So do we enable both? Or just one of them?

In separate communication, you said about an error for MSSQL@Azure: `NoSuchElementException: could not connect to any socket in [40.68.36.154:5986, 100.89.232.98:5986]`. That is the https port, so I assumed we were using https.

aledsage:
@neykov any comments from you about the need for http and https when we're using `Invoke-Command -Authentication CredSSP`?

guide/yaml/winrm/index.md
```diff
@@ -289,11 +289,25 @@ When a script is run over WinRM, the credentials under which the script are run
...
+- Up and running WinRM over http. Notice that we support winrm over https but for Invoke-Command to work it needs up and running winrm over http.
+  Apache Brooklyn can winrm over https but if the install script has inside it `Invoke-Command -Authentication CredSSP` then winrm over http has to be enabled as well.
+- Added trusted host entries which will use Invoke-Command
+- Allowed CredSSP
+
+All the above requirements are enabled in Apache Brooklyn through [brooklyn-server/software/base/src/main/resources/org/apache/brooklyn/software/base/custom-enable-credssp.ps1](https://github.com/apache/brooklyn-server/blob/master/software/base/src/main/resources/org/apache/brooklyn/software/base/custom-enable-credssp.ps1)
+script which enables executing commands with CredSSP in the general case.
+The script works for most of the Windows images out there version 2008 and later.
+
+Please ensure that Brooklyn's changes are compatible with your organisation's security policy.
```
ahgittin: [nit] US English `s/s/z/`

ahgittin:
useful.  @neykov if you have any comments on http(s) (re aled's comment above) it can be a new PR

neykov:
@aledsage Invoke-Command will default to http transport (the authentication protocol doesn't matter). The script can override it, but then needs to know what settings to use (what's available on the machine).
So the solution is to either enable http for loopback or somehow figure out what's available and use that. 

@bostko  can you confirm

Also the self signed certificate might be a problem for `Invoke-Command` - should be fine for localhost where it's already trusted, but for other cases need to use `Enter-PSSession` beforehand with ignore option.

bostko:
@neykov indeed. The script does what you wrote.
It enables winrm and adds machine's ip and localhost to trusted remote hosts.

PR #38, opened 2016-03-28
Additional getting started tabs

drigodwin:
New tabs for RPM and windows


murdoaird:
LGTM - makes the install options clearer

PR #49, opened 2016-04-22
Inclusion of catalog items inside other catalog items

neykov:


neykov:
Goes together with https://github.com/apache/brooklyn-server/pull/124

geomacy:
My thought on this is that the behaviour we are describing here, and I don't just mean the newly added stuff in this PR, is complicated enough that we should expand the text to at least include examples of each of the possibilities.  At minimum I'd say the new material needs illustration.

E.g.

Entries can be URL to another catalog file to include, inheriting the meta from the current hierarchy. For example:
```
brooklyn.catalog:
  displayName: Foo
  brooklyn.libraries:
  - http://some.server.or.other/path/my.jar
  items:
  - classpath:my-catalog-entries.bom
  - id: use-from-my-catalog
    item:
      type: some-type-defined-in-my-catalog-entries
```
In this case a jar file is added as a library, and then catalog items are added from the "my-catalog-entries.bom" file it contains, followed by an additional item that makes use of the definitions in the included file.

And similarly for the include case.  We should cover the various supported scenarios.

A question - do we really need the "include" keyword? Does the use of a URL as the item not do the same thing?  If we can avoid "include" it might avoid confusion as to whether to use it or not.

neykov:
+1 for an example, will add one.

`include` **must** be used when the context is a map - if you have sibling properties. It **may** be used if the link is the only thing in the item description. In your example adding an `include:` will work as well, turning the list item from a string to a map with a single entry.

geomacy:
Worth adding this as a comment in the docs.

neykov:
will do

ahgittin:
@neykov the comments above are nice to have but can be a follow-on PR.  this is a good addition already so merging.

neykov:
@ahgittin can you have a look at the related apache/brooklyn-server#124 PR as well. The docs here are incorrect without it merged.

PR #50, opened 2016-04-26
 Adds --bindAddress flag to server-cli-reference.md

f377isB:
Adds --bindAddress launch flag which was not documented elsewhere.

### DIFF
--- a/guide/ops/server-cli-reference.md
+++ b/guide/ops/server-cli-reference.md
@@ -74,6 +74,11 @@ This means that, once the task history is large, Brooklyn will continually use t
 memory. It will only expunge tasks from memory when this space is required for other objects within the
 Brooklyn process.
 
+### Web Console Bind Address
+
+To specify a local interface other than loopback (127.0.0.1) for the web console to bind to:
+
+    --bindAddress <IP>
 
 ## Configuration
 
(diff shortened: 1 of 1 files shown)


guide/ops/server-cli-reference.md
```diff
@@ -74,6 +74,11 @@ This means that, once the task history is large, Brooklyn will continually use t
 memory. It will only expunge tasks from memory when this space is required for other objects within the
 Brooklyn process.
 
+### Web Console Bind Address
+
+To specify a local interface other than loopback (127.0.0.1) for the web console to bind to:
```
neykov: The web console will by default bind to `0.0.0.0`. It's restricted to `127.0.0.1` only if `--noConsoleSecurity` is used.
The IP can also be set to `127.0.0.1` so suggest to change the wording here.

f377isB:
Agree - was unaware of the default bind so thank you.

Modified wording to:

The web console will by default bind to 0.0.0.0. It's restricted to 127.0.0.1 if the --noConsoleSecurity flag is used. To specify a local interface, or use the local loopback (127.0.0.1), for the web console to bind to you should use:

neykov:
Thanks, merging.

PR #51, opened 2016-05-02
Added a notice for RPMs in snapshot releases

drigodwin:
RPMs are not available in snapshot releases, this is now reflected with a notice and redirection.


neykov:
For context - rpm snapshots are not available because the apache build infrastructure doesn't have the `rpm` executable installed on slaves. The solution is to move the build to a docker container which we have control over.
LGTM.

PR #53, opened 2016-05-04
tidy #1 of maven build instructions

ahgittin:
however will likely move soon to default being to skip deb, rpm, and go-client, unless you opt *in*


alrick87:
LGTM

PR #55, opened 2016-05-06
Update the tests documentation

tbouron:
It explicitly says that the target for a test can be omitted if wrapped into a TestCase that already defines the target.

Here is how it looks like:
test-entities

### DIFF
--- a/guide/yaml/test/test-entities.md
+++ b/guide/yaml/test/test-entities.md
@@ -75,6 +75,10 @@ The `TestSensor` entity performs an assertion on a specified sensors value.
 - `timeout` - duration to wait on assertion to return a result. For example `10s`, `10m`, etc
 - `assert` - assertion to perform on the specified sensor value. See section on assertions below.
 
+<div class="alert alert-info">
+    <strong>Tip:</strong> If the <code>TestSensor</code> is wrapped within a <code>TestCase</code>, <code>ParallelTestCase</code> or <code>LoopOverGroupMembersTestCase</code> that set the target, <strong>you don't need to specify the target</strong>, unless you want to test another entity.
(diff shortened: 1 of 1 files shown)


guide/yaml/test/test-entities.md
```diff
@@ -89,6 +93,10 @@ The `TestEffector` entity invokes the specified effector on a target entity. If
 - `params` - parameters to pass to the effector, these will depend on the entity and effector being tested. The example above shows the `url` and `targetName` parameters being passed to Tomcats `deploy` effector.
 - `assert` - assertion to perform on the returned result. See section on assertions below.
 
+<div class="alert alert-info">
```
ahgittin: this would be a good candidate for a partial/include following the pattern of `_includes/feature-item.html` as used in `website//learnmore/features/*.md` -- you could just write

    {% include tip-testcase.html title="TestEffector" %}

also curious why you're using html rather than md syntax

but both comments are pedantic, feel free to ignore

merging

tbouron: The `include` seems nice but the line will become quite long depending on the text to pass right?

As for using HTML over md, I didn't know I could use markdown inside HTML so quite happy to change if you prefer.

ahgittin:
no, not at all worth refactoring, just wanted to let you know

at some point it would be worth refactoring this page so it has sub-menu entries and isn't one long scroll but it's eminently usable which is the main thing for now

as for longer text in templates, completely agree ... one day wouldn't it be nice to have integrated w angular or reach, to be able to define `<alert>` and then `<tip-testcases>` so we could inline just write `<tip-testcases type="TestEffector">`

PR #66, opened 2016-05-18
Windows - Advices for blueprints that are using restart

bostko:


guide/yaml/winrm/index.md
```diff
@@ -271,6 +271,16 @@ config like `pre.install.reboot.required` and `install.reboot.required`. If requ
 installation commands can be split between the pre-install, install and post-install phases
 in order to do a reboot at the appropriate point of the VM setup.
 
+We Strongly recommend to **write blueprints in a way that they do NOT restart automatically windows**
+
+Using install commands that restart the Operating System automatically through a forked process
+will make the blueprint to behave inconsistently and report wrongly failure.
+
+To avoid such cases manually call the command `shutdown /r` or the powershell one `Restart-Computer`.
```
aledsage: See the config option "install.reboot.required". That handles reboot better than executing this command directly.

aledsage:
LGTM. We could elaborate to say why not (i.e. say what the user can expect to happen if they do let the reboot happen). I'll merge as-is.

PR #70, opened 2016-05-22
Improve multi-location docs

aledsage:
In particular, how it can be used with the “availability zone” extension
to round-robin across the targets.


mikezaccardo:
LGTM -- this is documentation that would have been very helpful when I was developing a clustered blueprint targeted at multiple locations with round-robin desired, so it's great to add.

sjcorbett:
+1. @aledsage can you also open this as a pull request against 0.9.0 so it appears on the /latest section of the website?

PR #74, opened 2016-05-31
$brooklyn:component(ID) is deprecated, remove it from docs

neykov:


aledsage:
@neykov Makes sense, but in truth I'm not sure why `DslComponent.component(String)` is deprecated. The method `component(String scope, String id)` is not deprecated, and there is no equivalent `entity(String scope, String id)`.

My gut feel is that the problem is in `DslComponent`, and that we should add `entity(String, String)` and deprecate that other `component()` method.

So merging here.

PR #75, opened 2016-06-06
BROOKLYN-286: docs for entity-config merging

aledsage:


neykov:
LGTM, worth mentioning the special treatment of `templateOptions` for locations.

aledsage:
Thanks @neykov - I've added a section on `templateOptions`. Merging now.

PR #76, opened 2016-06-07
Docs/rename test ssh command

aledsage:


geomacy:
👍 looks good to me, merge away as far as I see it.

PR #80, opened 2016-06-23
Document shell.env json serialization behaviour.

neykov:
Depends on https://github.com/apache/brooklyn-server/pull/216

### DIFF
--- a/guide/yaml/custom-entities.md
+++ b/guide/yaml/custom-entities.md
@@ -82,6 +82,18 @@ to be specified up one level, alongside the type.
 However config keys which are *not* declared on the type *must* be declared in the `brooklyn.config` block. 
 
 
+### Passing custom variables
+
+Blueprint scripts can be parametrised through environment variables, making them reusable in different use-cases.
+Define the variables in the `env` block and then reference them using the standard bash notation:
+
+{% highlight yaml %}
+{% readj example_yaml/vanilla-bash-netcat-env.yaml %}
+{% endhighlight %}
+
+Non-string objects in the `env` map will be serialized to JSON before passing them to the script.
+
+
 #### Declaring New Config Keys
 
 We can define config keys to be presented to the user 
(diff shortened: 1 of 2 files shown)


guide/yaml/example_yaml/vanilla-bash-netcat-env.yaml
```diff
@@ -0,0 +1,12 @@
...
+location: localhost
+services:
+- type: org.apache.brooklyn.entity.software.base.VanillaSoftwareProcess
+  name: Simple Netcat Server
+  launch.command: |
+    echo $MESSAGE | nc -l $NETCAT_PORT &
+    echo $! > $PID_FILE
+    
+  env:
+    MESSAGE: hello
```
geomacy: maybe worth adding a $brooklyn.config(....) in as the value, to emphasise the reusability mentioned in the .md, here and in the port.

guide/yaml/custom-entities.md
```diff
@@ -82,6 +82,18 @@ to be specified up one level, alongside the type.
 However config keys which are *not* declared on the type *must* be declared in the `brooklyn.config` block. 
 
 
+### Passing custom variables
+
+Blueprint scripts can be parametrized through environment variables, making them reusable in different use-cases.
```
geomacy: typo "parametrised"

geomacy:
Looks good; short but sweet.  A couple of minor comments.

guide/yaml/custom-entities.md
```diff
@@ -82,6 +82,18 @@ to be specified up one level, alongside the type.
...
+Define the variables in the `env` block and then reference them using the standard bash notation:
```
geomacy: or 'shell.env' (is there any difference? what's preferred?)

neykov: Used env to be consistent with the other examples on the page. "env" is used when not using "brooklyn.config". "shell.env" is used when nested in "brooklyn.config" (that's the original idea, currently they are more or less interchangeable).

We should be consistently using "brooklyn.config" but that's for another PR.

guide/yaml/example_yaml/vanilla-bash-netcat-env.yaml
neykov: Parameters and DSL are introduced in the next section. Want to keep it short and simple and up to the point here.

neykov:
Thanks @geomacy - fixed the typo and merging.

PR #87, opened 2016-07-06
Fixed broken links

drigodwin:
Fixed broken links throughout the docs

### DIFF
--- a/_plugins/brooklyn_metadata.rb
+++ b/_plugins/brooklyn_metadata.rb
@@ -44,7 +44,7 @@ module BrooklynMetadata
         }
       end
       
-      url_set["git"] = "https://github.com/apache/brooklyn/tree/#{ git_branch }"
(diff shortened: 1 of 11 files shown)


_plugins/brooklyn_metadata.rb
```diff
@@ -44,7 +44,7 @@ def generate(site)
         }
       end
       
-      url_set["git"] = "https://github.com/apache/brooklyn/tree/#{ git_branch }"
+      url_set["git"] = "https://github.com/apache/brooklyn-server/tree/#{ git_branch }"
```
ahgittin: this could be wrong for some classes -- we need a more sophisticated strategy to determine if it's `server` or `library` etc (but this is probably good enough for now)

drigodwin: Yes, it needs re-working. For this though, I manually checked all the places this is being used and they're all currently in brooklyn-server.

PR #88, opened 2016-07-12
Add details about using wrong date

bostko:


guide/ops/troubleshooting/deployment.md
```diff
@@ -30,6 +30,11 @@ means there was some problem obtaining or connecting to the machine.
 
 An error like `... Not authorized to access cloud ...` usually means the wrong identity/credential was used.
 
+AWS requires a X-Amz-Date header which contains the date of the Apache Brooklyn AWS client.
+If the date on the server is wrong for example several hours behind you will get Authorization Exception.
```
neykov: Suggest changing to "is wrong even by a few minutes.."

andreaturli: I suspect this may apply to other cloud API as well. We should probably recommend to run Brooklyn on a machine with NTP properly configured?

PR #97, opened 2016-07-25
Improves examples copy&paste-ability

neykov:
Remove `$` from line beginnings, so that the example can easily be copy&pasted in the console.
Also fix tomcat port so that the example works out of the box.

### DIFF
--- a/guide/start/policies.md
+++ b/guide/start/policies.md
@@ -355,8 +355,8 @@ Tomcat on the vagrant VMs named "byon1" to "byon4":
 
 {% highlight bash %}
 $ for i in byon{1..4}; do
-$   vagrant ssh ${i} --command 'ps aux | grep -i tomcat |  grep -v grep | awk '\''{print $2}'\'' | xargs kill -9'
-$ done
+    vagrant ssh ${i} --command 'ps aux | grep -i tomcat |  grep -v grep | awk '\''{print $2}'\'' | xargs kill -9'
+  done
 {% endhighlight %}
 
 You can view the state of the Tomcat server with the command below (which drills into the  
@@ -435,14 +435,14 @@ you could use a load generator like jmeter, or use a script such as the one show
 (changing URL for the URL of your load-balancer):
 
 {% highlight bash %}
-$ URL=http://10.10.10.101:8000/
+$ URL=http://10.10.10.101:8080/
 $ for i in {1..600}; do
-$   for j in {1..50}; do 
-$     curl -saefafefj  ${URL} > /dev/null || echo "Curl failed with exit code $?"
-$   done
-$   echo "Finished batch $i"
-$   sleep 1
-$ done
+    for j in {1..50}; do
+      curl -saefafefj  ${URL} > /dev/null || echo "Curl failed with exit code $?"
+    done
+    echo "Finished batch $i"
+    sleep 1
+  done
 {% endhighlight %}
 
 While those curl commands run in a separate terminal, you can look at the metrics for the first


drigodwin:
LGTM :thumbsup:

guide/start/policies.md
```diff
@@ -435,14 +435,14 @@ you could use a load generator like jmeter, or use a script such as the one show
 (changing URL for the URL of your load-balancer):
 
 {% highlight bash %}
-$ URL=http://10.10.10.101:8000/
+$ URL=http://10.10.10.101:8080/
```
aledsage: I thought nginx would come up on 8000, rather than 8080 (which is the port that the first tomcat would use).

neykov: Good point, will revert.

guide/start/policies.md
```diff
@@ -355,8 +355,8 @@ Tomcat on the vagrant VMs named "byon1" to "byon4":
 
 {% highlight bash %}
 $ for i in byon{1..4}; do
-$   vagrant ssh ${i} --command 'ps aux | grep -i tomcat |  grep -v grep | awk '\''{print $2}'\'' | xargs kill -9'
```
aledsage: Where is our magic that used to treat "$" prefix in a special way (putting it into a separate div, so it wouldn't be included if someone highlighted the text to copy-paste it)? I can't find it.

@drigodwin said that he thinks it may have broken at around the same time as someone did a bootstrap upgrade.

@ahgittin @sjcorbett do you know where that code is?

sjcorbett: I don't.

ahgittin: clipboard support is working fine for me -- just click the icon

would be nice to exclude it from a manual copy/paste but that's harder.  gonna experiment with http://stackoverflow.com/questions/22159929/html-exclude-text-from-clipboard-copy

ahgittin:
i'd prefer a clever solution to avoiding the `$` rather than removing it from the display.  it is used throughout to distinguish what is entered and multi-line from what is returned.

neykov:
The change is not removing all `$`s - it's leaving the one at the command start. The removed characters are in the following lines in a multi-line command. 
See here for an example of what its trying to fix http://brooklyn.apache.org/v/latest/start/policies.html#auto-scaling.
The example snippet looks confusing with all those `$` on each line (that's a multi-line command). Messing with copy-pasteability is an additional problem.

ahgittin:
ah of course.  should we render it as:

```
$ for i in byon{1..4}; do
>   vagrant ssh ${i} --command 'ps aux | grep -i tomcat |  grep -v grep | awk '\''{print $2}'\'' | xargs kill -9'
> done
```

?

or i'd say to remove the leading char altogether

neykov:
I'm for the latter so it's easier to just grab it from the snippet and paste it somewhere. Otherwise you'd have the same problem as now - need to paste it somewhere intermediate to remove all those extra characters.

ahgittin:
@neykov if you use the clipboard icon in the corner it does that for you, though i agree if you don't know about it it's irritating -- trying to fix that

neykov:
The clipboard icon doesn't work on flash-free browsers so better not to depend on it doing the stripping.

ahgittin:
+1

ahgittin:
@neykov is https://github.com/apache/brooklyn-docs/pull/123 a better approach?
no strong feelings if you prefer to remove the `> ` multi-line prompt IE layer this on top of that

i suspect the `-saefafefj` flags to `curl` is a mistake which should be removed (and grep for it, i've noticed it in another file too) but want to confirm just in case it's a magic incantation on another `curl` impl /cc @grkvlt @Graeme-Miller

neykov:
Closing in favour of #123.
Good spot on the `curl` argument agree should be safe to remove.

PR #99, opened 2016-08-03
Move entropy/ulimit docs to troubleshooting

aledsage:
They were previously in website/documentation/increase-entropy.md and
in the faq. Now they are top-level pages in troubleshooting.

### DIFF
--- a/guide/ops/troubleshooting/increase-system-resource-limits.md
+++ b/guide/ops/troubleshooting/increase-system-resource-limits.md
@@ -1,0 +1,29 @@
+---
+layout: website-normal
+title: Increase System Resource Limits
+toc: /guide/toc.json
+---
+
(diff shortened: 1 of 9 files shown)


guide/ops/troubleshooting/increase-system-resource-limits.md
```diff
@@ -0,0 +1,30 @@
+---
+layout: website-normal
+title: Increase System Resource Limits
+toc: /guide/toc.json
+---
+
+You could encounter the error below when running with many entities:
```
drigodwin: I would rephrase to "If you encounter the error..." or "You may encounter the error..."

drigodwin:
Looks good! thanks @aledsage

aledsage:
Thanks @drigodwin - I've made that change. Merging now.

PR #100, opened 2016-08-05
Updated the maven archetype doc

Graeme-Miller:
The doc will now more explicitly show adding the JAR to Brooklyn


aledsage:
LGTM; merging.

PR #117, opened 2016-10-21
Winrm4j client diagnostic hints

bostko:


guide/yaml/winrm/client.md
```diff
@@ -0,0 +1,125 @@
...
+        IPv4Filter = *
+        IPv6Filter = *
+        EnableCompatibilityHttpListener = false
+        EnableCompatibilityHttpsListener = false
+        CertificateThumbprint
+        AllowRemoteAccess = true
+
+Use an Apache Brooklyn BYON blueprint to try easily other connection options.
+
+    location:
```
bostko: I may add that they users test commands preliminary with BYON.
Such as administrative commands which require CredSSP

PR #119, opened 2016-10-28
Opening ports for Security Group based locations

bostko:


guide/ops/locations/customizing_location_ports.md
```diff
@@ -0,0 +1,34 @@
+---
+title: Customizing Location Ports
+layout: website-normal
+---
+
+## Opening ports during startup.
+
+Use the `org.apache.brooklyn.location.jclouds.networking.SharedLocationSecurityGroupCustomizer` location customizer.
```
aledsage: I think these docs need more context, and caveats.

For example, the `SharedLocationSecurityGroupCustomizer` only works in clouds that use the security group abstraction (so not in Azure, SoftLayer, etc).

aledsage:
Where config keys define the explicit ports, these will be opened automatically. It's worth referencing other parts of the docs for this - see https://github.com/apache/brooklyn-docs/blob/master/guide/yaml/custom-entities.md#port-inferencing

In fact, do you think these new docs should be added to that section? Or that section extracted to its own page? I'm not sure, as this new page does talk specifically about configuring the location (though maybe it shouldn't do that!).

guide/ops/locations/customizing_location_ports.md
```diff
@@ -0,0 +1,34 @@
...
+    name: ports @ AWS
+    location:
+      jclouds:aws-ec2:us-west-2:
+        customizers:
+        - $brooklyn:object:
+            type: org.apache.brooklyn.location.jclouds.networking.SharedLocationSecurityGroupCustomizer
+            object.fields: {tcpPortRanges: ["900-910", "915", "22"], udpPortRanges: ["100","200-300"], cidr: "82.40.153.101/24"}
+        - $brooklyn:object:
+            type: org.apache.brooklyn.location.jclouds.networking.SharedLocationSecurityGroupCustomizer
+            object.fields: {tcpPortRanges: ["900-910", "915", "22"], udpPortRanges: ["100","200-300"], cidr: "10.40.153.101/16"}
```
aledsage: Perhaps just show the one example `SharedLocationSecurityGroupCustomizer`. The duplication (to open it to the second cird, with all the same ports) looks cumbersome and is probably not the most common pattern.

guide/ops/locations/customizing_location_ports.md
```diff
@@ -0,0 +1,34 @@
...
+    services:
+    - type: org.apache.brooklyn.entity.software.base.EmptySoftwareProcess
+
+Make sure that you have rule which makes port 22 accessible from Apache Brooklyn.
+
+## Opening ports during runtime.
+
+Example configuration in effector
```
aledsage: Need some additional explanation to say what this does - that it adds an effector with a given name, etc.

tbouron:
I think having separated docs for customising locations is fine. But I would definitely add the port-opened explicitly if we add configkeys, as mentioned by @aledsage.

Otherwise, LGTM

ahgittin:
@bostko good stuff here, are you able to incorporate the comments ?

bostko:
Hi, I rebased PR and addressed Aleds' comments for `SharedLocationSecurityGroupCustomizer`.
@aledsage , regarding the section talking about configuring the location, if using `provisioning.properties` 
 as you suggested then makes it location non specific.
My driver for splitting into two sections: location customizers is to have better visibility about what both do and document easily classes like `SudoTtyFixingCustomizer`, `AbstractEbsVolumeCustomizer` in jclouds location customizers section.

PR #121, opened 2016-11-07
document change to default coercion behaviour

ahgittin:
As tested in https://github.com/apache/brooklyn-server/pull/411

The failure to coerce a default value is a bug, but the fix might catch people out who used `$brooklyn:config` for ports in places without specifying custom config so worth adding to release notes.


aledsage:
LGTM; merging. Explanation feels complicated, but not obvious how to simplify it!

PR #123, opened 2016-11-07
Remove bash prompt chars

ahgittin:
inspired by #97 (and replacing it) this uses the proper multi-line bash char, and ensures these chars are removed on copy-paste

uses a nice JS `oncopy` trick mentioned in the comments on that PR

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -23,10 +23,8 @@ under the License.
 
 {{ content }}
 
-
-<script src="{{ site.path.style }}/js/zeroclipboard/ZeroClipboard.min.js"></script>
-
 <script language="JavaScript" type="application/javascript">
+
     fix_padding_function = function () { 
         $('body').css('padding-top', parseInt($('#header').css("height"))+10);
         $('body').css('padding-bottom', parseInt($('#footer').css("height"))+10);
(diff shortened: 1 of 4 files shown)


_layouts/base.html
```diff
@@ -65,32 +63,70 @@
...
-      // at the very start, or immediately following any newline). Correctly handles continuation
-      // lines, where a leading '$' or '%' is *not* a prompt character.
-      txt = txt.replace(/(^|[^\\]\n)[$%] /g, "$1");
-    }
+    var target0 = $(this).next();
+    var target = target0.clone();
+    target.find('.nocopy').remove();
+    var txt = target.text();
     clipboard.on( 'dataRequested', function (client, args) {
+      handleHideCopy( target0.closest('div.highlight') );  //not necessary but nicer feedback
```
neykov: This won't be called if zeroclipboard is not active (no flash installed for example). Could use [`user-select`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) for a similar effect.

ahgittin: yeah the `oncopy` handler solves the no-flash case.  would be nice if we didn't even show the clipboard on no-flash but didn't do that.

i have added `user-select` in a follow-up PR, but note it doesn't always do what one might expect, so this code is worth having in addition

neykov: Oh it's attached to the `oncopy` event - missed that. Should be good enough then.

neykov:
LGTM

PR #124, opened 2016-11-08
tidy curl args

ahgittin:
as discussed in #97 -- the old args looked really strange, at least on the os x `curl`.

would appreciate feedback from any of @drigodwin @duncangrant @Graeme-Miller @aledsage @grkvlt who might have some insight as to whether the old args were there for a reason.


drigodwin:
I agree, the `saefafefj` looks like a typo - LGTM

neykov:
Thanks, merging.

PR #126, opened 2016-11-17
Use brooklyn user for easier copy&paste

neykov:
The brooklyn user is the most common one used for running instances (created by the rpm & deb packages).


geomacy:
LGTM 👍 , merging.

PR #132, opened 2016-12-12
 Using Bundle-SymbolicName in Catalog

bostko:


guide/ops/catalog/index.md
```diff
@@ -149,6 +149,12 @@ The following optional catalog metadata is supported:
   and other metadata (such as versions, etc) may not be applied.
 - `brooklyn.libraries`: a list of pointers to OSGi bundles required for the catalog item.
   This can be omitted if blueprints are pure YAML and everything required is included in the classpath and catalog.
+  Each pointer to an OSGi bundle can be a string representing the JAR file url or a map.
+  A map consists of:
+    - `url: http://some.server.or.other/path/my.jar` Resource where JAR file is retrieved . Usually it is with prefix `http://` or ``file://`
+    - `name: myJar-BundleSymbolicName` If not supplied here it will be obtained from <br>`Bundle-SymbolicName` written in `my.jar:META-INF/MANIFEST.MF`
+    - `version: 1.0` If not supplied here it will be obtained from <br>`Bundle-Version` written in `my.jar:META-INF/MANIFEST.MF`
```
neykov: Don't include this section. There's no good reason why one would need the expanded variant.

guide/ops/catalog/index.md
```diff
@@ -161,6 +167,30 @@ The following optional catalog metadata is supported:
...
+  - id: use-from-my-catalog
+    version: "1.0.0"
+    itemType: entity
+    iconUrl: classpath://myJar-BundleSymbolicName:org/apache/brooklyn/MyEntity/icon.png
+    item:
+      type: myJar-BundleSymbolicName:org.apache.brooklyn.MyEntity
+      brooklyn.config:
+        some.config: "some value"
+        templates.install:
+          classpath://myJar-BundleSymbolicName:org/apache/brooklyn/MyEntity/config.xml: config.xml
```
neykov: We should try to fix cases where the prefix is needed. Does the example work without the prefixes? If not can you file a bug against Brooklyn.

guide/ops/catalog/index.md
```diff
@@ -120,14 +120,14 @@ brooklyn.catalog:
...
+  - classpath://myJar-BundleSymbolicName:my-catalog-entries-inside-jar.bom
   - some-property: value
-    include: classpath://more-catalog-entries-inside-jar.bom
+    include: classpath://myJar-BundleSymbolicName:more-catalog-entries-inside-jar.bom
   - id: use-from-my-catalog
     version: "1.0.0"
     itemType: entity
     item:
-      type: some-type-defined-in-my-catalog-entries
+      type: myJar-BundleSymbolicName:some-type-defined-in-my-catalog-entries
```
neykov: Same as previous comment, prefixes shouldn't be required. Actually should be discouraged and used as a last resort to get things working.
File bugs against Brooklyn if you find cases where they are needed.

neykov:
Would prefer to include usage of the prefix in a troubleshooting section or similar, not as the go to way of using types/resources.

guide/ops/catalog/index.md
ahgittin: i've just put together a wee test in https://github.com/apache/brooklyn-server/pull/485 ... classpath resolution seems good for the most part

so tidy this line, and lines 181 and 185

(though we may need a fix to make 185 work, using the right `ResourceUtils` to load the `iconUrl`)

ahgittin: actually `iconUrl` should work fine, looks like `CatalogResourceTest` tests this

bostko:
Bundle-Symbolic name should be discouraged. PR is irrelevant.

PR #133, opened 2016-12-19
fixed mis-spelling by removing Broo(o)klyn

murdoaird:
Page reads just as well without Brooklyn ref.


drigodwin:
Thanks @murdoaird, merging now

PR #135, opened 2016-12-24
Link to lists.a.o instead of mail-archives.a.o

ajs6f:
lists.apache.org is perhaps a bit nicer and more functional than mail-archives.apache.org.


aledsage:
Thanks @ajs6f - LGTM. Merging.

PR #138, opened 2017-01-17
Use full configkey name in docs rather than the shorter version

tbouron:


tbouron:
This follow [the proposal of deprecating the `@SetFromFlag` annotation](http://mail-archives.apache.org/mod_mbox/brooklyn-dev/201611.mbox/browser)

PR #143, opened 2017-01-26
Update disk usage explanation

bostko:


guide/ops/requirements.md
```diff
@@ -12,7 +12,7 @@ The size of server required by Brooklyn depends on the amount of activity. This
 * the amount of management and monitoring required per entity
 
 For dev/test or when there are only a handful of VMs being managed, a small VM is sufficient.
-For example, an AWS m3.medium with one vCPU, 3.75GiB RAM and 4GB disk.
+For example, an AWS m3.medium with one vCPU, 6GiB RAM and 8GB disk.
```
drigodwin: The previous values were correct for `m3.medium` see: https://aws.amazon.com/ec2/instance-types/#m3

drigodwin:
This change is only true for the Karaf distibution of Apache Brooklyn. This is not the default so I would suggest holding off merging this PR until it is.

guide/ops/requirements.md
```diff
@@ -24,8 +24,8 @@ any binaries for custom blueprints/integrations.
 There are three main consumers of disk space:
 
 * **Static files**: these are the Apache Brooklyn distribution with its own
-  dependencies, plus binaries for custom blueprints and integrations added to
-  the `lib` directory. Note that Brooklyn requires that Java is installed which
+  dependencies, plus OSGi bundles for custom blueprints and integrations added to
+  the `data/cache` directory. Note that Brooklyn requires that Java is installed which
```
neykov: Custom bundles would be added to `deploy` folder, `data/cache` is an internal detail.

tbouron:
Now that Brooklyn uses Karaf as default launcher, we should probably merge this @drigodwin

PR #144, opened 2017-01-26
Configuring logging from karaf distribution of Apache Brooklyn

bostko:


guide/dev/tips/logging.md
```diff
@@ -11,6 +11,10 @@ with a default configuration which just works out of the box
 and bindings to the other common libraries (``java.util.logging``, ``log4j``, ...)
 if you prefer one of those.
 
+Modify `etc/org.ops4j.pax.logging.cfg`.
```
geomacy: Needs to be in a separate section for Karaf.  Karaf log control is not based on logback. And in any case with Karaf you would be better to explain about use of Karaf console to set log levels (use `log:set` command).  The only time you would need to modify the file by hand would be if for some reason it needed modified before the first start of Karaf.

tbouron:
@bostko Can you update the PR as per as @geomacy's comment?

guide/dev/tips/logging.md
bostko: @geomacy to what extend logging is configurable for OSGi distribution through config files in `brooklyn-logback-includes` module?

bostko:
@tbouron I included @geomacy note.

PR #147, opened 2017-01-27
Added known issues section to azure arm

Graeme-Miller:


guide/ops/locations/_azure-ARM.md
```diff
@@ -94,3 +94,9 @@ The loginUser can be anything, as long as it's specified.
 
 The `overrideAuthenticateSudo: true` key tells Apache Brooklyn that default on Azure images do not have passwordless sudo 
 configured by default.
+
+#### Known issues
+There are currently two know issues with Azure ARM:
```
drigodwin: know -> known

guide/ops/locations/_azure-ARM.md
```diff
@@ -94,3 +94,9 @@ The loginUser can be anything, as long as it's specified.
...
+
+* It can take a long time for VMs to be provisioned
+* Azure ARM appears to have some fairly strict rate limiting
```
drigodwin: I would mention these are limits on the API usages rather than the number of machines and cause `AzureComputeRateLimitExceededException`s

tbouron: I would also mention that the rate limit is shared between classic and ARM

PR #148, opened 2017-02-09
Adds doc for urlEncode DSL

aledsage:


guide/yaml/yaml-reference.md
```diff
@@ -213,6 +213,9 @@ concise DSL defined here:
   `$brooklyn:root()`.
 * `$brooklyn:formatString("pattern e.g. %s %s", "field 1", "field 2")` returns a future which creates the formatted string
   with the given parameters, where parameters may be strings *or* other tasks such as `attributeWhenReady`
+* `$brooklyn:urlEncode("val")` returns a future which creates a string with the characters escaped
+  so it is a valid part of a URL. The parameter can be a string *or* another task. For example,
+  `$brooklyn:urlEncode($brooklyn:config(\"mykey\"))`
```
neykov: Let me repeat my concerns about this functionality here - see https://github.com/apache/brooklyn-server/pull/551#issuecomment-278158142.
At the very least can you document that this is meant only for query parameters. While not perfect it can do the job until improved. The various parts of the URL have different rules for escaping so this won't work universally.

guide/yaml/yaml-reference.md
```diff
@@ -213,6 +213,9 @@ concise DSL defined here:
   `$brooklyn:root()`.
 * `$brooklyn:formatString("pattern e.g. %s %s", "field 1", "field 2")` returns a future which creates the formatted string
   with the given parameters, where parameters may be strings *or* other tasks such as `attributeWhenReady`
+* `$brooklyn:urlEncode("val")` returns a future which creates a string with the characters escaped
```
drigodwin: I'm not sure it's worth using a term like 'future' here which the user may not be aware of. From their perspective this function just URL Encodes the parameter?

guide/yaml/yaml-reference.md
aledsage: Agreed - I've added:
```
It uses "www-form-urlencoded" for the encoding, which is appropriate for query 
parameters but not for some other parts of the URL (e.g. space is encoded as '+').
```

guide/yaml/yaml-reference.md
aledsage: @drigodwin agreed - the wording I used here is consistent with the rest of the descriptions. I'll leave it as-is for this PR. We should revisit to rewrite the descriptions of the entire "DSL Commands" section. An introductory paragraph that says about how the DSL is resolved is probably worth while (it's important that users understand it's a future for any non-trivial use-case, so they can understand when/if it will block (and also what happens if you compose DSL).

---
We are also missing a better description and examples of how to compose the DSL - e.g. are we recommending what to me feels like "idiomatic yaml" like:
```
$brooklyn:formatString:
- "http://%s" 
- $brooklyn:attributeWhenReady("host.subnet.hostname")
```

versus `$brooklyn:formatString("http://%s", $brooklyn:attributeWhenReady("host.subnet.hostname"))`, which looks more like a programming language and less like configuration. (I get that it is effectively programming, but that doesn't mean the user needs to write it like that!).

aledsage:
@neykov's comment addressed; @drigodwin deferred for a future PR. Merging now.

PR #150, opened 2017-02-16
Add deprecated functionality which can't be communicated to the user otherwise

neykov:
Can't be communicated to the user by deprecating APIs or logging.

Follow up to https://github.com/apache/brooklyn-server/pull/562.


tbouron:
LGTM

PR #152, opened 2017-03-07
Docs for SharedLocationSecurityGroupCustomizer

iyovcheva:


drigodwin:
Thanks @iyovcheva, I think it would be worth specifying that this only works on some clouds and perhaps listing the clouds.

iyovcheva:
Thanks @drigodwin, added a note for that.

PR #156, opened 2017-03-14
Suggest using unique custom.machine.namer.machine

bostko:
Notes on using unique VM name to avoid duplicated names in cloud.

### DIFF
--- a/guide/ops/locations/_clouds.md
+++ b/guide/ops/locations/_clouds.md
@@ -278,7 +278,10 @@ The name that Apache Brooklyn generates for your virtual machine will, by defaul
 If you have created a lot of virtual machines, this kind of naming may not be helpful. This can be changed using the following YAML in your location's `brooklyn.config`:
 
     cloudMachineNamer: org.apache.brooklyn.core.location.cloud.names.CustomMachineNamer
-    custom.machine.namer.machine: My-Custom-Name-${entity.displayName}
+    custom.machine.namer.machine: My-Custom-Name-${entity.displayName[0..*10]}-${entity.id}
+
+_Note that often Cloud Provider require Virtual Machine name to be unique.<br>
+Some jclouds providers use that machine node name for other cloud resources such as Security Groups._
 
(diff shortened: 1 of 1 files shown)


aledsage:
Instead of this change, I think it would be better to add to `guide/ops/locations/_azure-ARM.md`, to say that the cloud-machine-namer needs to give a unique name for each VM (which is different from other clouds)! In aws for example, jclouds appends some randomly generated suffix for the security group it creates (I believe).

If I'm right about that difference, then @bostko can you please open an issue in jclouds so that it behaves the same as other clouds for appending a random name (cc @andreaturli).

See https://docs.microsoft.com/en-us/azure/architecture/best-practices/naming-conventions for azure's guidelines on naming.

guide/ops/locations/_clouds.md
```diff
@@ -278,7 +278,10 @@ The name that Apache Brooklyn generates for your virtual machine will, by defaul
 If you have created a lot of virtual machines, this kind of naming may not be helpful. This can be changed using the following YAML in your location's `brooklyn.config`:
```
aledsage: I think we should change the wording in this doc. I think it's wrong to say "this kind of naming may not be helpful". The point is more that:

    However, some organization have their own naming convention for VMs. The naming strategy used by
    Brooklyn can be configured using the following YAML in your location's `brooklyn.config`:

If you have created a lot of virtual machines, this kind of naming may not be helpful. This can be changed using the following YAML in your location's brooklyn.config:

ahgittin:
@bostko ping

bostko:
@aledsage I think problem I had was that I used machine location namer which was specifiying a constant machine name. Thus I hit an error in Azure ARM where the first problem was when jclouds was creating resource group since it got the a name which already existed.
@aledsage thanks for naming convention suggestion. It seems good to have `-rg` suffix for resource group names, `-sg` suffix for security group names, however I do not find the need for it at the moment.
Closing the PR for now.

PR #157, opened 2017-03-15
Adds “DSL Evaluation of Inherited Config” section [WIP]

aledsage:
_I'm about to create the bug report, and will update the text to reference that!_


aledsage:
Thanks @drigodwin - I've added a bug report and fixed the link (to https://issues.apache.org/jira/browse/BROOKLYN-455). Will merge this now.

PR #162, opened 2017-04-06
BROOKLYN-460: Brooklyn Camp syntax guidance

bostko:


guide/yaml/yaml-reference.md
```diff
@@ -79,38 +79,44 @@ the entity being defined, with these being the most common:
...
+  Entities, policies, and initializers may accept additional key-value pairs,
+  usually documented in their documentation (e.g. javadoc), or in the case of Java
+  often as static fields in the underlying Java class.
+  Often there are config keys or flags (indicated by `@SetFromFlag`) declared on the class;
+  these declared flags and config keys may be passed in at the root of the `ServiceSpecification` or in `brooklyn.config`.
+  (Undeclared config is only accepted in the `brooklyn.config` map.)
+  Referencing the parameters from within java classes is identical to using config keys. In yaml it's
+  usually referenced using `$brooklyn:scopeRoot().config("displayName")`. See below for more details on scopes.
+
+* `brooklyn.tags`: documents list of tag objects which should be assigned to the entity.<br/>
```
drigodwin: documents **a** list of tag objects

drigodwin: I also do not think there should be `<br />`s throughout this.

guide/yaml/yaml-reference.md
```diff
@@ -79,38 +79,44 @@ the entity being defined, with these being the most common:
...
+  A tag can be an object which is not of type `org.apache.brooklyn.util.core.task.DeferredSupplier`.<br/>
```
drigodwin: What does this mean? Why is it important it's not of that specific type?

guide/yaml/yaml-reference.md
```diff
@@ -79,38 +79,44 @@ the entity being defined, with these being the most common:
...
+  **Tags must be simple objects** which can be serialized in persistence state and when retrieved by the REST API.<br/>
```
drigodwin: persisted state

guide/yaml/yaml-reference.md
geomacy: Better to explain from a user's point of view, i.e. say that only some DSL is allowed, where values can be resolved immediately, i.e. basically can't use `attributeWhenReady` or anything with the same `whenReady` semantics.

PR #166, opened 2017-04-11
Remove snapshot docs link

drigodwin:
Snapshot Versions links are defunct and not useful (see: [BROOKLYN-443](https://issues.apache.org/jira/browse/BROOKLYN-443)). I propose we remove them.


aledsage:
I agree. We must not encourage users to use a snapshot build, via the official docs. (We're not doing that - there are no snapshot download links, but as such the snapshot docs aren't useful for end-users.)

Merging.

PR #171, opened 2017-04-19
A ConfigKey default value should be immutable.

bostko:


bostko:
Closing in favor of #171

PR #172, opened 2017-04-19
Java config docs: default-value should be immutable

aledsage:


nakomis:
I think it would be worth changing the wording to make it clearer that you're not talking about the user *setting* the default to a new value, but instead modifying the underlying object, i.e. the risk isn't that they might say `myConfigKey.setDefaultValue(newValue)`, but rather `myListConfigKey.getDefaultValue().add(foo)`. Perhaps something like:

If supplying an object as a default value, it is important that the object is immutable. Otherwise, it risks users of the blueprint modifying the underlying object, which would affect blueprints that are subsequently deployed.

drigodwin:
Thanks for the addition @aledsage, do you think this should replace @bostko's #171?

geomacy:
LGTM, merging.

PR #174, opened 2017-04-26
Security guidelines

bostko:
Ssh access should be strongly limited.

### DIFF
--- a/guide/ops/security-guidelines.md
+++ b/guide/ops/security-guidelines.md
@@ -72,7 +72,8 @@ root login.
 ### VM Users
 
 It is strongly discouraged to use the root user on VMs being created or managed by Brooklyn.
-
+SSH-ing on the VM should be done on rare cases such as initial Apache Brooklyn setup,
+Apache Brooklyn upgrade and other important maintenance occasions.
 
 ### SSH keys
 


guide/ops/security-guidelines.md
```diff
@@ -72,7 +72,10 @@ root login.
 ### VM Users
 
 It is strongly discouraged to use the root user on VMs being created or managed by Brooklyn.
+SSH access should be delegated on rare cases such as initial Apache Brooklyn setup and other maintenance occasions.
```
drigodwin: I'm not sure what you mean by delegated here?

guide/ops/security-guidelines.md
```diff
@@ -72,7 +72,10 @@ root login.
...
 
+Avoid putting config a lot of options in `etc/brooklyn.cfg` which one would later need to ssh and change.
+Keep location configuration it is preferable to use [Locations in the Catalog](./catalog/#locations-in-the-catalog).
```
drigodwin: I think these two lines need rewording but I'm not sure what they're adding. `etc/brooklyn.cfg` is a config file in the Karaf Brooklyn distro which is not currently the default. I'm not sure we're advising people put locations in there anywhere.

PR #182, opened 2017-05-12
Enforce the use of `brooklyn.config` in yaml examples

tbouron:
This is related to the proposal of deprecating `@SetFromFlag` annotation, see [email thread](https://lists.apache.org/thread.html/e36eebe022ca0304332f9d66fd5b3203672febaf5e39af4707ab9e8a@%3Cdev.brooklyn.apache.org%3E).

It enforces the use of `brooklyn.config` as well as the real configkey names. With this, the documentation becomes consistent.

### DIFF
--- a/guide/blueprints/chef/advanced-chef-integration.md
+++ b/guide/blueprints/chef/advanced-chef-integration.md
@@ -32,13 +32,14 @@ A general schema for the supported YAML is below:
 
 {% highlight yaml %}
 - type: chef:cookbook_name
-  cookbook_urls:
-    cookbook_name: url://for/cookbook.tgz
-    dependency1: url://for/dependency1.tgz
-  launch_run_list: [ "cookbook_name::start" ]
-  launch_attributes: # map of arguments to set in the chef node
-  service_name: cookbook_service
-  pid_file: /var/run/cookbook.pid
+  brooklyn.config:
+    brooklyn.chef.cookbooksUrls:
+      cookbook_name: url://for/cookbook.tgz
+      dependency1: url://for/dependency1.tgz
+    brooklyn.chef.runList: [ "cookbook_name::start" ]
+    brooklyn.chef.launch.attributes: # map of arguments to set in the chef node
+    brooklyn.chef.lifecycle.serviceName: cookbook_service
+    brooklyn.chef.lifecycle.pidFile: /var/run/cookbook.pid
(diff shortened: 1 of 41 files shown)


aledsage:
@tbouron I agree in principle with this change, however I have a big reservation about doing this now.

First, my agreements:
* We should be consistent in our examples and in what we support. We should always use `brooklyn.config` (or if not, then we should have a separate proposal to never use it).
* We should have one authoritative name for the each config key, and always use that in our examples. We should deprecate the other aliases.

Now for my disagreement... Once we have support for deprecating config key names, we should carefully decide which name we want to be the authoritative name (is the the current "real" name, or is the alias actually a better name; or is there a new name we should change it to that is more consistent with the rest of Brooklyn!)

Therefore I worry that this PR will change our examples, and then we'll change some of those back again when we figure out what the best names are. As an example, the `AutoScalerPolicy`'s config key should be `metric` rather than `autoscaler.metric` - i.e. the name used in the alias should be switched to be the authoritative name. Having the prefix "autoscaler" is just needless repetition.

What makes the long-term choice of naming a bit tricky is our support for runtime-management config inheritance, i.e. inheriting config keys from a parent entity (*). Where a user does that, we want to reduce the risk of name collisions. But frankly I think we want to discourage it!

(*) Our current behaviour is horrible - a config value that uses the alias is not picked up the child entities, but a config key that uses the "real name" is picked up. However, users find it extremely hard to tell the difference between "real name" and "alias" when they're looking at examples, so invariably find this behaviour baffling.

aledsage:
At the risk of making more work for us @tbouron, perhaps we should split this PR into three parts (where 1 and 2 could be done together):
1. change example structure to be consistent (e.g. always using `brooklyn.config`).
2. change examples that use a bad-looking alias, where the alternative "real name" would make that particular example look cleaner/simpler.
3. after we support deprecation and we've changed each name to be the "best" name, we update the examples to use those authoritative names.

What I want to avoid is us temporarily changing the docs such that the examples look more complicated, in ways that we end up changing back again in not very long.

tbouron:
Doing (1) and (2) sounds good, I'll revert the real names by flags when it's too long/complicated.

tbouron:
@aledsage I reverted examples to the flag names where this was simpler and not likely to clash due to config inheritance

PR #183, opened 2017-05-17
[For Review] Details on configuring logging for OSGi distribution

bostko:
I am looking for some hints which we can add in docs for logging configuration.
Should we describe karaf commands in docs?

I also would like to clear the confusion between pid file config and xml config.
I think xml config is not used, right?
If so obviously we can delete https://github.com/apache/brooklyn-server/blob/0.11.x/karaf/features/src/main/feature/feature.xml#L46


bostko:
Any suggestions are welcome.

bostko:
Squash changes into https://github.com/apache/brooklyn-docs/pull/144

PR #185, opened 2017-05-19
[Suggestion] Merge guide/dev into website/developers

bostko:
One section for all coding related matters.


website/developers/index.md
```diff
@@ -1,14 +1,21 @@
 ---
 layout: website-normal
 title: Developers
-started-pdf-exclude: true
```
ahgittin: think this might be required for PDF build?

guide/ops/osgi.md
```diff
@@ -65,6 +65,8 @@ JAVA_DEBUG_OPTS='-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5
 For other options please check your JVM JPDA documentation.
 Hotspot JPDA:  https://docs.oracle.com/javase/8/docs/technotes/guides/jpda/
 
+## Quick testing
```
ahgittin: Either this should have content or next section should be a subsection (`###` not `##`)

ahgittin:
Do not merge yet.  In addition to the comments the case for making this change needs to be made.

The reason for the old docs structure is that some of the dev guide content is specific to a given version.  We lose that clarity with this PR.  I expect @bostko is suggesting that precision is not worth the user-facing complexity, but @bostko it would be helpful to describe how this is simpler and note main places confusion could arise if we don't have the dev guide aligned for released versions.

nakomis:
@ahgittin Can you take a look at this please, and see if you're happy with the changes @bostko has made

PR #190, opened 2017-05-23
Fix recommended OS / Java version

aledsage:


guide/ops/requirements.md
```diff
@@ -64,14 +64,14 @@ rotation according to your organisation's policy.
...
+The recommended operating system is CentOS 7.x or RedHat 7.x.
 
-Brooklyn has also been tested on Ubuntu 14.04 and OS X.
+Brooklyn has also been tested on CentOS 6.x, Ubuntu 14.04, Ubuntu 16.04 and OS X.
 
 
 ## Software Requirements
 
-Brooklyn requires Java (JRE or JDK) minimum version 1.7. 
+Brooklyn requires Java (JRE or JDK) minimum version 1.8.
```
neykov: Let's switch to the major version convention - Java 8?

guide/ops/requirements.md
```diff
@@ -64,14 +64,14 @@ rotation according to your organisation's policy.
 
 ## Supported Operating Systems
 
-The recommended operating system is CentOS 6.x or RedHat 6.x.
+The recommended operating system is CentOS 7.x or RedHat 7.x.
 
-Brooklyn has also been tested on Ubuntu 14.04 and OS X.
+Brooklyn has also been tested on CentOS 6.x, Ubuntu 14.04, Ubuntu 16.04 and OS X.
```
bostko: I would also add Windows here.

aledsage: Interesting. I think we need three levels then: recommented; tested; and "some folk also use...". We don't want to promise that it will work on all these things if we don't test it on all of them thoroughly every release. Only a subset of the testing is done by people running Windows.

ahgittin:
See #190 where this is merged and comments addressed.

I went with a simpler OS message as I think @bostko is right, we do test on and support Windows.  It does without saying some OS's will be tested more than others.  If we discover shortcomings in any supported OS we should treat that as a bug with appropriate prioritisation.

PR #202, opened 2017-06-30
fixing broken links

murdoaird:


guide/ops/osgi.md
```diff
@@ -29,7 +29,7 @@ Start the instance with a console in the foreground using the following command
 bin/karaf
 {% endhighlight %}
 
-This will launch the [Karaf console](https://karaf.apache.org/manual/latest/users-guide/console.html)
+This will launch the [Karaf console](http://karaf.apache.org/manual/latest/)
```
drigodwin: Should this be https://karaf.apache.org/manual/latest/#_using_the_console

murdoaird: Possibly, I was trying to ensure we linked to a higher level so our are links are more durable & require less maintenance @drigodwin

drigodwin: That is a good plan, a broken anchor would still load the page however and we should use https where possible.

PR #204, opened 2017-08-01
Fix wrong indentation in Windows Azure ARM config

bostko:


ygy:
LGTM

guide/locations/_azure-ARM.md
```diff
@@ -158,9 +158,8 @@ brooklyn.catalog:
       destroyOnFailure: false
 
       templateOptions:
-      overrideLoginUser: brooklyn
-      overrideLoginPassword: "secretPass1!"
-      resourceGroup: <ResourceGroup-name>
```
drigodwin: Did you mean to remove this line @bostko?

bostko: Yes, for Linux Azure ARM locations Apache Brooklyn docs does not advice to use a specific `resourceGroup`. I do not see a reason why we should advice users to use fixed resourceGroup.
I tested without this parameter and it works.

PR #205, opened 2017-08-07
use ~~~ rather than ```

Graeme-Miller:
This keeps this page in line with other pages. Also it was causing issues in a downstream project.


m4rkmckenna:
@Graeme-Miller Can you change the other places where ``` is used also

tbouron:
Good catch, kramdown uses `~` instead of `

LGTM

m4rkmckenna:
@tbouron @Graeme-Miller https://kramdown.gettalong.org/parser/gfm.html

Graeme-Miller:
@m4rkmckenna updated

PR #214, opened 2017-09-19
Document `sshMonitoring.enabled`

aledsage:
And some other tidies of custom-entities doc.

These docs are primarily for the change in https://github.com/apache/brooklyn-server/pull/824. It also documents how to wire up an alternative http-based polling for health.

IT also does some other tidy up. For example, `./start.sh` is no longer the default value for `launch.command` so we need to fix that example.

### DIFF
--- a/guide/blueprints/custom-entities.md
+++ b/guide/blueprints/custom-entities.md
@@ -46,12 +46,47 @@ before running `launch.command` relative to where that file is installed (or unp
 with the default `launch.command` being `./start.sh`.
 
 So if we create a file `/tmp/netcat-server.tgz` containing just `start.sh` in the root
-which consists of the two lines in the previous example,
+which contains the line `echo hello | nc -l 4321`, 
 we can instead write our example as: 
 
 {% highlight yaml %}
 {% readj example_yaml/vanilla-bash-netcat-file.yaml %}
 {% endhighlight %}
+
+
+#### Determining Successful Launch
+
+The default method used to determine a successful launch of `VanillaSoftwareProcess` is to run a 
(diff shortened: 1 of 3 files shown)


guide/blueprints/custom-entities.md
```diff
@@ -46,14 +46,48 @@ before running `launch.command` relative to where that file is installed (or unp
...
 we can instead write our example as: 
 
 {% highlight yaml %}
 {% readj example_yaml/vanilla-bash-netcat-file.yaml %}
 {% endhighlight %}
 
 
+#### Determining Successful Launch
+
+The default behaviour for `VanillaSoftwareProcess` is to run a command over ssh to do a
```
drigodwin: The default method used to determine a successful launch of `VanillaSoftwareProcess`

guide/blueprints/custom-entities.md
```diff
@@ -46,14 +46,48 @@ before running `launch.command` relative to where that file is installed (or unp
...
 {% endhighlight %}
 
 
+#### Determining Successful Launch
+
+The default behaviour for `VanillaSoftwareProcess` is to run a command over ssh to do a 
+health check. The health check is done post-launch (repeating until it succeeds, before 
+then reporting that the entity has started).
+
+The default command will check if the pid written to `$PID_FILE` is running. This is why
```
drigodwin: The default command used to carry out this health check will determine if the pid, written to `$PID_FILE` is running.

aledsage:
Thanks @drigodwin - comments incorporated and squashed into one commit. Merging now.

PR #215, opened 2017-09-20
Update Azure ARM documentation for Azure CLI 2.0

rdowner:
The existing instructions are for Azure CLI 1.0. Searching for "Azure CLI" returns hits for 2.0 so it's likely our users will end up with CLI 2.0.


andreaturli:
thanks @rdowner, lgtm

PR #222, opened 2017-10-16
Migrate docs to Gitbook

tbouron:
As per as the discussion on the ML: https://lists.apache.org/thread.html/760a3e2fdefaff8d8ac4ea3f98a45060fbaa0d886fa57c8f44e4742e@%3Cdev.brooklyn.apache.org%3E


tbouron:
Thanks @m4rkmckenna ! 

However, can you hold off a wee bit more? I just realised that while internal relative links are perfectly fine for the docs, it might cause issues for downstream projects therefore I would like to update those links to use the same trick we did before, i.e. `{{book.url}}/relative/path/to/root.md`.

I'll ping you once it's done

tbouron:
@m4rkmckenna Done, should be good to go now

m4rkmckenna:
Thanks @tbouron merging now

PR #225, opened 2017-11-01
Use gitbook-plugin-blueprint-tour to avoid rendering issue

tbouron:
I noticed that during the build, the rendering of pages could sometimes be wrong if markdown syntax is wrapped around HTML code in `*.md` (or the other way around) Note that it's not always the case. It looks like Gitbook is a bit flaky on this particular topic.

So I created a small plugin [`gitbook-plugin-blueprint-tour`](https://github.com/tbouron/gitbook-plugin-blueprint-tour) to encapsulate the rendering of the blueprint tour. Not only it fixes the rendering issue, but it is also cleaner IMHO as we don't mix HTML within markdown.


nakomis:
LGTM

tbouron:
@nakomis Hum I need to review the work I did on this one. Probably safe to ignore it for now, would that work for you @nakomis ?

PR #228, opened 2017-11-02
add docs on upgrades esp new blueprint bundle headers (beyond the release notes and javadoc)

ahgittin:
updates to release notes following recent changes to bundles, including https://github.com/apache/brooklyn-server/pull/874


ahgittin:
got toc structure working so added that here, closing #229

guide/ops/upgrades/index.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Upgrades
+---
```
tbouron: Front-matter is not needed anymore with gitbook

ahgittin: @tbouron i simply moved `guide/ops/upgrade.md` to this file.  useful comment but rather than block this PR because of a piecemeal improvement to existing content probably we should merge and have another PR remove unnecessary front-matter from all files?

guide/ops/upgrades/_blueprints.md
```diff
@@ -0,0 +1,50 @@
...
+
+To add these definitions, use the following headers in the bundle's OSGi `META-INF/MANIFEST.MF`:
+
+* `Brooklyn-Catalog-Force-Remove-Bundles`
+* `Brooklyn-Catalog-Force-Remove-Legacy-Items`
+* `Brooklyn-Catalog-Upgrade-For-Bundles`
+* `Brooklyn-Catalog-Upgrade-For-Types`
+
+The most common patterns are to indicate that a bundle can replace all previous versions of itself
+and all types thereing with types in the current bundle of the same name, using:
```
aledsage: typo: `therin`

guide/ops/upgrades/_blueprints.md
```diff
@@ -0,0 +1,50 @@
...
+```
+Brooklyn-Catalog-Force-Remove-Bundles: *
+```
+
+The above items can also take a range syntax, e.g. `"*:[1,2)"` when releasing a `2.0.0` to restrict to
+versions equal to or greater than `1.0.0` but less than `2.0.0`. (Note that ranges must be quoted.)
+Entries can also take comma-separated lists, and in the case of replacements, they can define
+explicit renamed targets using `sourceNameAndVersionRanges=targetNameAndVersion` entries.   
+These fields are defined in full in the
+[`BundleUpgradeParser`'s javadoc]({{book.url.brooklyn_javadoc}}/org/apache/brooklyn/core/typereg/BundleUpgradeParser.html).
```
aledsage: This is good enough for now, but longer term I think the docs should be the definitive source of info for this (particularly as we want to target non-java developers). I guess to limit duplication, the javadoc could point coders at the docs.

That whole topic needs more thought and discussion, for how we javadoc + document things!

guide/ops/upgrades/index.md
```diff
@@ -0,0 +1,22 @@
...
+
+This section provides all necessary information to upgrade Apache Brooklyn for both the RPM/DEB and Tarball packages.
+
+## Backwards Compatibility
+
+Apache Brooklyn version 0.12.0 onward runs primarily inside a Karaf container. When upgrading from 0.11.0 or below,
```
aledsage: (Not necessarily for this PR, but worth noting: we should update this wording for 1.0.0 because we'll only support the Karaf container (rather than just "primarily")).

ahgittin:
TY @aledsage .  Agree on all comments.  Have fixed typo and merged.

Re javadoc, there is something very handy about comments in javadoc but agree it is inelegant referring people to it.  I wonder if there is any way we could say in our docs to "grab javadoc on Class or C?lass.method and insert here" ?  cc @tbouron

tbouron:
Javadoc tool generates only an HTML per class so you can potentially include this in gitbook (probably needs a custom plugin)
I don't see however how to include only a specific method

PR #229, opened 2017-11-02
[HELP] Upgrades - restructure

ahgittin:
builds on #228 attempting to put things into a better structure

but gitbooks makes my pages non-linkable.  what's going on?!  cc @tbouron @rdowner


ahgittin:
got working, included in #228 now

PR #236, opened 2017-11-14
Refer to inheritance.runtime instead of deprecated inheritance.runtime

bostko:
Some useful change which was sitting for sometime in my workspace.


aledsage:
LGTM, merging; thanks @bostko!

PR #237, opened 2017-11-15
instructions for using a keypair in AWS

ahgittin:


guide/locations/_AWS.md
```diff
@@ -43,6 +43,51 @@ Below are examples of configuration options that use values specific to AWS EC2:
...
+      type: jclouds:aws-ec2
+      brooklyn.config:
+        identity: XXXXXXXXXXXXXXXX
+        credential: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXx
+  - id: aws-us-east-1
+    item:
+      type: aws-base
+      region: us-east-1
+      keyPair: XXXXXXXXX
+      login.privateKeyData: |
```
andreaturli: isn't `loginUser.privateKeyData`? there is also `loginUser.privateKeyFile` which can maybe be more convenient in some cases.

guide/locations/_AWS.md
```diff
@@ -43,6 +43,51 @@ Below are examples of configuration options that use values specific to AWS EC2:
   For example, `securityGroups: mygroup1` or `securityGroups: [ mygroup1, mygroup2 ]`.
 
 
+### Using a Registered Key Pair
+
+You can specify a `keyPair` to use for initial provisioning as a configuration option.
+If this is omitted, Brooklyn will use jclouds to create a new ad hoc key pair at AWS
+for that machine, and it will delete it afterwards.  This is usually seamless and
+occurs behind the scenes, with the post-provision user set up and configured as normal
+for all locations.  However if you are optimizing for creation of many machines,
```
andreaturli: [minor] I'm not sure the comment "However ..." is relevant here as it's not Brooklyn responsibility to prepare reliable images that work also when creating many machines at once. I'd remove it.

guide/locations/_AWS.md
ahgittin: you are completely right @andreaturli - so how the heck is my setup working with `login.privateKeyData` ???

andreaturli:
Maybe you are creating a new keypair without noticing?

guide/locations/_AWS.md
ahgittin: aha i have `brooklyn.location.jclouds.extraSshPublicKeyData` set so i think it is using `TemplateOptions.authorizePublicKey` and then using my normal `id_rsa`.  let me disable that and triple check these instructions are right.  TY!

ahgittin: no, it wasn't that, it's that config in location defns need to be in `brooklyn.config` -- updating this

guide/locations/_AWS.md
ahgittin: good point, have cleaned this up

ahgittin:
thanks @andreaturli - all fixed now

ahgittin:
scratch that.  it still doesn't seem to be installing my `keyPair`.

the logs say `importing` rather than `creating` the keypair (and complain if `keyPair` is set to something that doesn't exist) but the resulting image isn't sshable with the keypair :(

lemme try a different image, in case it's a problem with the default (rightimage)

ahgittin:
actually instructions here are fine.  was failing for me because i was also setting `extraSshPublicKeyData`

seems like the combo of `TemplateOptions.authorizePublicKey(String)` and `EC2TemplateOptions.keyPair(String)` makes jclouds fail

ahgittin:
good to merge now @andreaturli ?

(can sort out that combo problem separately)

andreaturli:
lgtm

Agreed combo problem can be dealt in a subsequent PR

Thanks!

PR #238, opened 2017-11-17
Specify EBS boot volume size

drigodwin:
Document how to specify the EBS boot volume size on AWS

### DIFF
--- a/guide/locations/_AWS.md
+++ b/guide/locations/_AWS.md
@@ -42,6 +42,22 @@ Below are examples of configuration options that use values specific to AWS EC2:
   [security groups](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html).
   For example, `securityGroups: mygroup1` or `securityGroups: [ mygroup1, mygroup2 ]`.
 
+### EBS boot volume size
+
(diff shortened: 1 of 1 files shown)


guide/locations/_AWS.md
```diff
@@ -42,6 +42,21 @@ Below are examples of configuration options that use values specific to AWS EC2:
...
+The size of the EBS boot volume for a VM can be specified using `mapNewVolumeToDeviceName` in `templateOptions` as shown below:
+
+    location:
+      jclouds:aws-ec2:
+        region: ...
+        templateOptions:
+          mapNewVolumeToDeviceName:
+           - /dev/sda1
+           - 123
+           - true
```
tbouron: Looks like the indentation is wrong here: should be either on the same level as`mapNewVolumeToDeviceName` or down by 2 spaces

guide/locations/_AWS.md
```diff
@@ -42,6 +42,21 @@ Below are examples of configuration options that use values specific to AWS EC2:
...
+
+Where `/dev/sda1` is the device name of the boot volume on the selected AMI and `123` is the size of the volume in gigabytes.
```
tbouron: Probably worth explaining what the 3rd parameters is?

guide/locations/_AWS.md
drigodwin: The indentation here is correct

PR #240, opened 2017-12-06
Race condition causing main.uri to be null

duncangrant:
Fixed by adding all required sensors to triggerSensors


geomacy:
looks good to me

PR #241, opened 2017-12-18
Quote included file name

sjcorbett:
Without the quotes `npm run book` fails cryptically: "template names must be a string: NaN"


sjcorbett:
@drigodwin would you mind checking this?

PR #256, opened 2018-06-04
BROOKLYN-588: add to troubleshooting for curl failure 

aledsage:
See https://issues.apache.org/jira/browse/BROOKLYN-588

### DIFF
--- a/guide/ops/troubleshooting/connectivity.md
+++ b/guide/ops/troubleshooting/connectivity.md
@@ -104,8 +104,8 @@ example, execute `sudo iptables -n --list` and `iptables -t nat -n --list`.
 ## Cloud firewalls
 Some clouds offer a firewall service, where ports need to be explicitly listed to be reachable.
 
-For example, [security groups for EC2-classic]
-(http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html#ec2-classic-security-groups)
+For example, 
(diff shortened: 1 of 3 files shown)


guide/ops/troubleshooting/connectivity.md
```diff
@@ -104,8 +104,8 @@ example, execute `sudo iptables -n --list` and `iptables -t nat -n --list`.
 ## Cloud firewalls
 Some clouds offer a firewall service, where ports need to be explicitly listed to be reachable.
 
-For example, [security groups for EC2-classic]
-(http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html#ec2-classic-security-groups)
+For example, 
+[security groups for EC2-classic](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-network-security.html)
```
tbouron: Why did you removed the anchor here? It is still valid.
Having said that, the section doesn't seem to have specificities to firewalls. So the title probably need to change to `security groups for EC2`, or be dropped all together

aledsage: Good spot. I removed the anchor because it's not just ec2-classic that is used in Brooklyn (one can point it at VPC security groups). Therefore I've changed the link text to 'security groups for AWS EC2'.

PR #258, opened 2018-06-20
Fix copy-state usage for brooklyn-karaf

aledsage:
In brooklyn classic, one could run `brooklyn copy-state ...`. This is not supported in brooklyn karaf.

Instead, the REST api can be used. This PR updates the docs to show how to use it.

Please first review/merge https://github.com/apache/brooklyn-server/pull/972, to fix this REST api for usage in karaf!

PR #260, opened 2018-06-21
Use the official AWS CLI instead of s3cmd

tbouron:
Based on this comment: https://github.com/apache/brooklyn-docs/pull/258#discussion_r197133399

Note: I checked the docs and this is the only place where we suggested to use `s3cmd`


nakomis:
LGTM

duncangrant:
Merging - thanks @tbouron

PR #263, opened 2018-07-26
Update docs to match the new Brooklyn UI

tbouron:
This is the required changes to match the newly donated Brooklyn UI (see https://lists.apache.org/thread.html/20d82876f7e84244c2c4d3464a1531a10af50ca3d20e9e1de97a5faf@%3Cdev.brooklyn.apache.org%3E)

It updated the relevant portions of text as well as all the screenshot.

Note that another PR is required for the website part: https://github.com/apache/brooklyn-docs/pull/264


ahgittin:
Fabulously thorough job.

I think `guide/ops/troubleshooting/web-console-issues.md` is still relevant, so we shouldn't delete that page, should we?  (Also, if we _do_ delete it, shouldn't there be a reference to that page which is deleted -- I didn't see that, and note `guide/SUMMARY.md` still refers to it.)

There is also a reference to `BrooklynJavascriptGuiLauncher` on `./dev/env/ide/index.md`.  As per comment https://github.com/apache/brooklyn-server/pull/974#issuecomment-408083341 this is no longer supported and we should identify the best way in each IDE to run with the UI.

However that can be deferred, and the docs at this stage in this section could simply link to a REST CLI launcher class and note that we would welcome suggestions on the best way to run with osgi and the UI from within IDEs.  Or if you know how to do OSGi+UI launches, add those notes, but do keep the instructions that Brooklyn+REST can be launched non-OSGi, maybe noting that some features around bundle management will be unavailable.

tbouron:
> Fabulously thorough job.

Thanks!

> I think guide/ops/troubleshooting/web-console-issues.md is still relevant, so we shouldn't delete that page, should we? (Also, if we do delete it, shouldn't there be a reference to that page which is deleted -- I didn't see that, and note guide/SUMMARY.md still refers to it.)

Well, we never saw this issue with the new UI hence why I deleted the page. I don't think this is relevant anymore. But thanks for the nudge about `SUMMARY.md`, I forgot to check the references, will remove.

> There is also a reference to BrooklynJavascriptGuiLauncher on ./dev/env/ide/index.md. As per comment apache/brooklyn-server#974 (comment) this is no longer supported and we should identify the best way in each IDE to run with the UI.

Didn't remove the reference as I don't think it should be part of this set of PRs (commented this on the relevant PR) But yeah, this can be updated in a following PR.

tbouron:
@ahgittin Added back the web console issues page + updated the IDE setup

PR #264, opened 2018-07-27
Update website to match the new Brooklyn UI

tbouron:
This updates the web console screenshot with the new UI


ahgittin:
there is a conflict but makes sense to me to apply the change intended here

PR #266, opened 2018-08-24
Clean up docs

kemitix:
* fix a few typos
* extract some code blocks to `example_yaml`
* fix loading bash script as tar.gz from url to ensure it is executable
* fix hi netcat example to close the connection at end of file
* wrap jdbc string format in quotes

### DIFF
--- a/guide/blueprints/example_yaml/vanilla-bash-netcat-w-client.yaml
+++ b/guide/blueprints/example_yaml/vanilla-bash-netcat-w-client.yaml
@@ -66,7 +66,11 @@ services:
       name: sayHiNetcat
       description: Echo a small hello string to the netcat entity
       command: |
-        echo $message | nc -N $TARGET_HOSTNAME 4321
+        # Uncomment the appropriate command for your operating system
+        # for Linux:
+        # echo $message | nc -N $TARGET_HOSTNAME 4321
+        # for MacOS:
+        # echo $message | nc $TARGET_HOSTNAME 4321
       parameters:
         message:
           description: The string to pass to netcat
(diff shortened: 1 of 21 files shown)


guide/blueprints/example_yaml/vanilla-bash-netcat-w-client.yaml
```diff
@@ -66,7 +66,7 @@ services:
       name: sayHiNetcat
       description: Echo a small hello string to the netcat entity
       command: |
-        echo $message | nc $TARGET_HOSTNAME 4321
+        echo $message | nc -N $TARGET_HOSTNAME 4321
```
tbouron: What the `-N` does? Also, I'm not sure this is advisable as some platform, i.e. MacOS don't have this option available

kemitix: Struggling to find how you would install an updated version of netcat on MacOS that would support this command. The problem is that on some platforms the `nc` command doesn't close the channel when `echo` terminates, which prevents the pipeline command from terminating.

Would a suitable solution be to include both options with one of them commented out and a comments saying `# MacOS` and `# Linux`?
e.g.
```yaml
# Uncomment the appropriate command for your operating system
# Linux
# echo $message | nc -N $TARGET_HOSTNAME 4321
# MacOS
# echo $message | nc $TARGET_HOSTNAME 4321
```

tbouron: Yeah, or just add a note at the bottom to say that for MacOS, you need to remove `-N` 👍

PR #276, opened 2019-01-30
Add note about including parameters when using 'wrap:' prefix

kemitix:
The `Bundle-SymbolicName` and `Bundle-Version` parameters need to be specified to avoid default values being generated that can cause problems when rebinding persisted state.

### DIFF
--- a/guide/blueprints/java/bundle-dependencies.md
+++ b/guide/blueprints/java/bundle-dependencies.md
@@ -127,8 +127,10 @@ there are a few options for getting an equivalent OSGi bundle:
 
 * Use the `wrap:` prefix. The [PAX URL Wrap protocol](https://ops4j1.jira.com/wiki/display/paxurl/Wrap+Protocol) 
   is an OSGi URL handler that can process your legacy jar at runtime and transform it into an OSGi bundle.  
-  This can be used when declaring a dependency in your feature.xml, and when using the Karaf console's 
-  `bundle:install`. Note that it is not yet supported in Brooklyn's `brooklyn.libraries` catalog items.
+  This can be used when declaring a dependency in your feature.xml, and when using the Karaf console's
+  `bundle:install`. Note that it is not yet supported in Brooklyn's `brooklyn.libraries` catalog items.__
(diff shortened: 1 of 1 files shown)


guide/blueprints/java/bundle-dependencies.md
```diff
@@ -127,8 +127,10 @@ there are a few options for getting an equivalent OSGi bundle:
 
 * Use the `wrap:` prefix. The [PAX URL Wrap protocol](https://ops4j1.jira.com/wiki/display/paxurl/Wrap+Protocol) 
   is an OSGi URL handler that can process your legacy jar at runtime and transform it into an OSGi bundle.  
-  This can be used when declaring a dependency in your feature.xml, and when using the Karaf console's 
-  `bundle:install`. Note that it is not yet supported in Brooklyn's `brooklyn.libraries` catalog items.
+  This can be used when declaring a dependency in your feature.xml, and when using the Karaf console's
+  `bundle:install`. Note that it is not yet supported in Brooklyn's `brooklyn.libraries` catalog items.__
```
ahgittin: why the `__` at the end of this line?

ahgittin:
good addition, clear text.  one minor comment.

guide/blueprints/java/bundle-dependencies.md
kemitix: It's markdown syntax for a hard line break.

ahgittin: Aha did not know.  What's the difference between that and just leaving a blank line after?

Merging.

kemitix: Actually, that should have been two spaces, not underscores. I misunderstood what my text editor was showing me. I'll change them.

kemitix: @ahgittin Didn't want to try leaving a blank line as it is within a list, and the markdown processor would likely break that into a full paragraph rather then a part of the list text.

ahgittin: I think they're pretty smart about intervening new lines as long as indentation makes sense.  But not sure and they vary widely.  Didn't know about the two-space trick either.  Will merge you #277 .

PR #280, opened 2019-03-19
Add text to comply with Apache naming guidelines

geomacy:
See http://www.apache.org/foundation/marks/pmcs.html#naming

In particular this part:
<quote>
Every product homepage and any overview download pages for the product
must include a prominent reference to the product as "Apache Foo
software", and must include a brief one sentence description of the
function of the software product itself.
</quote>

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -96,3 +96,9 @@ DEPENDENCIES
   jekyll
   json
   therubyracer
+
+RUBY VERSION
+   ruby 2.1.2p95
+
+BUNDLED WITH
+   1.16.6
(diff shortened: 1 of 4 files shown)


geomacy:
I've made some adjustments and moved the description down as suggested by @rdowner. On my box now the front page looks like

image


and the download page like 

image

What do you think?

geomacy:
hi @rdowner @tbouron could I ask for another review, what do you think of the look now?

tbouron:
I'm also happy with this @geomacy !

geomacy:
Thanks @rdowner @tbouron, I've merged this now and will have a look at pushing it to the website. Haven't done this before, we'll see how it goes. It will probably be the weekend though before I get to look at it...

PR #281, opened 2019-04-07
Update download index and verify files to match published website

geomacy:
This is a follow up to "Add text to comply with Apache naming guidelines" #280

When I went to push the updated docs to the website I discovered the source and the published website had drifted apart a bit on the download index page and the verify page.

This PR mainly brings the `website` branch back into line with the published site, but also makes another small modification to re-instate the `KEYS` file link, which should exist on the live site but doesn't, and tidies up the introductory text on the verify page a little bit.

There are probably some other things to fix, such as broken links, but that can be done in a separate PR.

### DIFF
--- a/download/index.md
+++ b/download/index.md
@@ -12,39 +12,24 @@ children:
  Apache Brooklyn is software for modelling, deploying and managing cloud applications through autonomic blueprints.
 </h5>
 
+## Version {{ site.brooklyn-stable-version }} (recommended)
+Version {{ site.brooklyn-stable-version }} is the current version recommended for production use.
+
 <div class="row">
 
 <div class="col-md-6" markdown="1">
 
 <div class="panel panel-default">
   <div class="panel-heading" markdown="1">
-#### RPM Package
+#### Packages
   </div>
   <div class="panel-body" markdown="1">
 <div style="height: 9em;" markdown="1">
-Suitable for version 7 of CentOS and Red Hat Enterprise Linux.
+RPM package for version 7 of CentOS and Red Hat Enterprise Linux, and a DEB package for Ubuntu and Debian distributions.
 </div>
 
 <div class="text-center">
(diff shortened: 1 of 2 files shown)


download/index.md
```diff
@@ -66,8 +146,8 @@ A pre-built package that contains Apache Brooklyn and all of its dependencies in
 </div>
 
 <div class="text-center">
-  <a class="btn btn-default" href="https://www.apache.org/dyn/closer.lua?action=download&filename=brooklyn/apache-brooklyn-{{ site.brooklyn-stable-version }}/apache-brooklyn-{{ site.brooklyn-stable-version }}-bin.tar.gz" role="button">Tarball</a>
-  <a class="btn btn-default" href="https://www.apache.org/dyn/closer.lua?action=download&filename=brooklyn/apache-brooklyn-{{ site.brooklyn-stable-version }}/apache-brooklyn-{{ site.brooklyn-stable-version }}-bin.zip" role="button">Zip</a>
+  <a class="btn btn-default" href="https://www.apache.org/dyn/closer.lua?action=download&filename=brooklyn/apache-brooklyn-1.0.0-M1/apache-brooklyn-1.0.0-M1-bin.tar.gz" role="button">Tarball</a>
```
tbouron: Shouldn't we use `{{site.brooklyn-stable-version}}` there? Same question for the remaining versions below

geomacy: The way this is structured is to have two sets of files, one at the top for 0.12.0 which is still the official recommended version, and then an equivalent set below for 1.0.0-M1. This is to match up with what is currently on https://brooklyn.apache.org/download/index.html. This line is just the start of the links for 1.0.0-M1.

geomacy: well actually they start a bit above

tbouron: Ahaa, I didn't realised there was 2 columns as I only scanned the code. Looks good then!

PR #285, opened 2019-04-16
Add Jenkinsfile to build and push automatically the website

tbouron:
As the title says. This should be picked up by https://builds.apache.org/view/B/view/Brooklyn/job/brooklyn-docs/ and build only for PRs, then build and publish for each commit in `website` branch


ahgittin:
cool - merging

PR #295, opened 2019-12-03
Updates localhost doc to add a deployable example

nakomis:
Also bumps some of the dependencies, which was required
to build the docs


README.md
```diff
@@ -36,7 +36,7 @@ Once the `node` and `npm` are installed, run this command to install all the req
 at the correct versions:
 
 ```bash
-npm install
+npm install; npm install natives
```
tbouron: There is a huge warning about this package: https://www.npmjs.com/package/natives

Better to go with the recommandation and install `graceful-fs` instead

package.json
```diff
@@ -3,7 +3,9 @@
   "version": "1.0.0",
   "description": "Apache Brooklyn Documentation",
   "main": "index.js",
-  "dependencies": {},
+  "dependencies": {
+    "natives": "^1.1.6"
```
tbouron: Should go into the `devDependencies`

nakomis:
@tbouron As discussed offline, `natives` is (unfortunately) still required as `graceful-fs` doesn't work for us. I have, however, moved it into `devDependencies` and removed the `npm install natives` from the README.md

PR #298, opened 2020-01-06
Added documentation snippet to show how entities can be tagged and added to Quick Launch

iuliana:
Added documentation snipped to show how entities can be tagged and added to the quick launch. 
The documentation snippet can be found in the `Catalog Items YAML Syntax`


iuliana:
Very nice idea @tbouron. Done. :)

PR #305, opened 2020-05-19
Split policies docs into sub-pages

aledsage:
(no change to text; just moves it around)


jcabrerizo:
LGTM @aledsage, thanks!

PR #309, opened 2020-09-11
Document recent addition to allow arbitrary beans to be added to catalog and used in config, initializers, and elsewhere

ahgittin:


geomacy (approved):
LGTM, one semi-typo.

One optional thought, perhaps it would be good to add an example spec using a bean.

ahgittin:
Thanks!  Fixed and merged.

PR #315, opened 2021-05-20
Switch from NodeJS/Gitbook back to Ruby/Jekyll, using latest Ruby

ahgittin:
This reverts the swtich to Gitbook, as nice as it was, it long ago stopped being nicely supported OSS so simplest thing to do is to go back to Ruby/Jekyll.  Unfortunately it has moved on and a lot of things here got broken -- but we think this fixes them all.


duncangrant (approved):
LGTM.  I've tested locally and I can set up jekyll and build following the instructions.  Pages seem to be generated correctly and internal links seem to work.  
It's a shame that gitbook has moved away from OSS but Jekyll is being actively developed and is currently well supported so this is a sensible move.
image

PR #316, opened 2021-06-01
add docs for composer annotations and hints

ahgittin:
docs for https://github.com/apache/brooklyn-ui/pull/210 (and some previous work)


guide/blueprints/yaml-reference.md
```diff
@@ -281,3 +281,43 @@ elements for more information:
...
+  services:
+  - type: server
+    brooklyn.tags:
+      - ui-composer-annotation: A simple default annotation
+      - ui-composer-annotation:
+          text: >-
+            Shown below, yellow text, centered with CSS. Because it's long, scroll bars horizontally and vertically shown when
+            needed.
+          styleInnerDiv: 'margin: auto; color: yellow;'
+          'y': 120
```
algairim, line 305: Do we need quotes for `y`?

ahgittin: not sure why, but the JS yaml processor puts these in.  probably it wants to treat `y` like a reserved word for `yes` aka `true`.  i just cut and pasted.  but agree, worth removing here.

PR #317, opened 2021-07-15
Ensure `baseurl` is correctly respected

ahgittin:
In all places where links are made (text, menu, styling).

To facilitate links in text, any of `target.md`, `./target.md`, `target.html`, `/path/to/target.md`, `/path/to/target.html` can be used, and the framework will recognize that as pointing at that target and will write to the appropriate html output, with `baseurl` prepended if there is a baseurl and the path is absolute


iuliana (approved):
Amazing. :)

PR #329, opened 2021-08-31
major tidy, use canonical approach to embedded blueprints

ahgittin:
blueprints are refactored into a `*.camp.{md,yaml}` convention, allowing these to be extracted for tests or augmented with other dialects.  also simplify paths and improve how files are imported (`read` liquid tag).


algairim (approved):

iuliana (approved):
LGTM

PR #333, opened 2021-09-02
Fix broken links

duncangrant:
When we deploy the brooklyn docs to different paths, e.g. for different versions, contents links that had a # were not correctly  generated.  This fixes this.

Also I cleaned up some docs to move yaml into separate files.


jcabrerizo (approved):
LGTM @duncangrant thanks

PR #355, opened 2022-06-06
Display top-level sections as chidlren only

algairim:
Signed-off-by: Mykola Mandra <mykola.mandra@cloudsoft.io>


algairim:
This is a proposal to tidy up children displayed in the docs under CLI ref guide.

PR #356, opened 2022-06-06
Do not replace relative links on the same page

algairim:
Signed-off-by: Mykola Mandra <mykola.mandra@cloudsoft.io>


algairim:
This change breaks absolute links when `baseurl` is used in docs build, reverted in https://github.com/apache/brooklyn-docs/pull/364.

PR #379, opened 2023-05-19
Workflow - update-children and more [WIP]

ahgittin:


jcabrerizo (approved):
Thanks @ahgittin , this sounds really powerful. Looking forward to see it running

guide/blueprints/workflow/nested-workflow.md
```diff
@@ -80,6 +119,8 @@ and always allowing 1.
 This might be used for example to upgrade a cluster in situ, leaving the larger of 10 instances or half the cluster alone, if possible.  
 If the concurrency expression evaluates to 0, or to a negative number whose absolute value is larger than the number of values, the step will fail before executing, to ensure that if e.g. "-10" is specified when there are fewer than 10 items in the target list, the workflow does not run.  (Use "max(1, -10)" to allow it to run 1 at a time if there are 10 or fewer.)
 
+Concurrency cannot be specified when `reducing`.
```
iuliana, line 122: @ahgittin  Maybe add a **Note:** in front of this statement

iuliana (changes requested):
One non-critical suggestion.

One critical: either the stack example is wrong, or we are not processing the `aws cloudformation describe-stacks` json right, because I am getting `InvalidReferenceException: The following has evaluated to null or missing:
==> describe_stacks_output_json.Stacks  [in template "workflow" at line 1, column 3]` when the last step is executed

iuliana (approved):

PR #380, opened 2023-05-25
Fixes incorrect filename in aws-cfn-stacks docs

nakomis:
Fixes an incorrectly referenced filename (aws-cfn-type.yaml -> aws-cfn-type.bom)


jcabrerizo (approved):
tks @nakomis

PR #391, opened 2026-01-22
add a winrm limitation section for using % character in env vars

zan-mateusz:
Add a section to WinRM limitations doc page.

`%` can't be used in env variables passed via WinRM, as it results in error: `Parameter is incorrect`.


iuliana (commented):
LGTM.

